PR #1, opened 2020-11-23
Fixed a typo on service name

iRhonin:


ilyazzz (approved):
In the future I'll try to make a proper crates.io package and/or distro packages instead of this script.

PR #13, opened 2021-03-02
Changed sockets to nix crate

ashleysmithgpu:
So we can use abstract sockets and not worry about file permissions or cleaning up the socket. Buffer read code may need some work

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,8 +44,6 @@ sudo systemctl enable --now lactd
 ```
 You can now use the application.
 
-Note: By default, only members of the group `wheel` can change settings. This can be changed by editing `/etc/lact.json` and changing the `group` value.
-
 # CLI
 
 There is also a cli available.
--- a/daemon/Cargo.toml
+++ b/daemon/Cargo.toml
@@ -17,3 +17,4 @@ rand = "0.8"
 signal-hook = "0.3"
 pciid-parser = { git = "https://github.com/ilyazzz/pci-id-parser.git" }
 reqwest = { version = "0.11", features = ["blocking", "json"] }
+nix = "0.20"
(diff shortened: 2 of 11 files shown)


ilyazzz:
This seems like a really nice thing to have, but I have some questions about how it works:

- As I understand, the buffer read loop will read from the socket until the buffer is filled. Is that correct? The current implementation uses `read_to_end` because the connection to the socket closes after an action has been sent, for the exact purpose of not having to worry about the buffer length or signaling that the message has ended. Or is this buffer size here only arbitrary and doesn't have to match the size of the message?

- I'm not too familiar with how exactly unix sockets work, does nix create a global named socket, or is there a socket file stored somehwere?

Otherwise I'm happy to merge this, this looks like a lot less hacky than calling `chown` in the shell.

ashleysmithgpu:
Hi,

> As I understand, the buffer read loop will read from the socket until the buffer is filled. Is that correct? The current implementation uses read_to_end because the connection to the socket closes after an action has been sent, for the exact purpose of not having to worry about the buffer length or signaling that the message has ended. Or is this buffer size here only arbitrary and doesn't have to match the size of the message?

I would have preferred to use read_to_end but I'm not sure it works with unix sockets. read() will apparently return 0 when there is nothing left to read, <0 on error and anything else is the size of what it just wrote to `buffer`. So as long as all the message sizes are less than `BUFFER_SIZE` it should be good. Basically it doesn't have to match exactly.

> I'm not too familiar with how exactly unix sockets work, does nix create a global named socket, or is there a socket file stored somehwere?

This is an "abstract socket" so it doesn't create a file to manage and we don't have to worry about [cleaning up the socket.](https://unix.stackexchange.com/questions/216784/does-linux-automatically-clean-up-abstract-domain-sockets) Unfortunately we can't use std::net because it doesn't support the null byte at the start of the path which denotes an abstract socket.

Anyway, I don't commit to git much but I was working on something similar and thought I'd have a look quickly :) Sorry for the sudden PR

ilyazzz:
The largest buffer I've seen is ~3000 so it should be fine for now.
I wasn't aware of abstract sockets. Thanks for info.
And PRs are always welcome.

PR #214, opened 2023-11-24
lact-gui: Use libadwaita

### LINKED ISSUE #213: App is using light theme when system is dark theme on Fedora 39

ZVNexus:
Fixes #213

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1257,6 +1257,7 @@ dependencies = [
  "gtk4",
  "lact-client",
  "lact-daemon",
+ "libadwaita",
  "pretty_assertions",
  "tracing",
  "tracing-subscriber",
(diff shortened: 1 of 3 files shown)


ilya-zlobintsev:
Using libadwaita is something I've considered during the initial migration from GTK3 to GTK4, and at the time I rejected the idea due to the fact that libadwaita applications do not follow system themes.

At the same time, I've since found that some things look broken with the breeze theme that KDE uses by default, so after all it might be worth to force a theme in order to ensure that things look correct. However, I'm still not entirely certain about this.

The dark theme issue is more of a problem with distro configuration. But it seems like most programs either use GTK4 togther with libadwaita, or GTK3, so the less common scenario of plain GTK4 might not have been considered/tested in the distro.

Conan-Kudo:
I would prefer not forcing adwaita. There are significant improvements to Breeze for GTK 4 that are coming as part of Plasma 6, done in large part by @pontaoski and @easyteacher. If there are still issues, [please report them to KDE](https://bugs.kde.org/enter_bug.cgi?product=Breeze&component=gtk%20theme) so we can resolve them.

ilya-zlobintsev:
After some consideration, I think it's better to leave theming up to the user/distro configuration. The original issue can be resolved by installing the right packages for a non-libadwaita dark theme, and I don't think it's worth forcing a theme everywhere simply to avoid it.

ilya-zlobintsev:
@Conan-Kudo I've reported the issue that I've hit with Breeze: https://bugs.kde.org/show_bug.cgi?id=477505

Conan-Kudo:
Thanks!

PR #217, opened 2023-11-25
fix: allow gui build without drm

pastalian:
lact-daemon with default features depends on drm, which breaks build without drm.


ilya-zlobintsev:
Thanks, I've overlooked this

PR #220, opened 2023-11-25
fix: Use DESTDIR properly in the Makefile and introduce PREFIX

Conan-Kudo:
This allows for packaging tooling to use the Makefile properly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,18 @@
 export CARGO_TARGET_DIR ?= ./target
-DESTDIR ?= /usr/local
+DESTDIR ?= /
+PREFIX ?= /usr/local
 
 build-release:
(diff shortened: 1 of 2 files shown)


ilya-zlobintsev:
Am I understanding this right: after this change, the install options for packaging will be: 
```
DESTDIR=${pkgdir} PREFIX=/usr make install
```
Also, the [packaging script](https://github.com/ilya-zlobintsev/LACT/blob/master/pkg/recipes/lact/recipe.yml#L38) needs to be updated.

Conan-Kudo:
Yes, and I just fixed the recipe file.

ilya-zlobintsev:
Thanks!

PR #221, opened 2023-11-25
fix: set lactd to run without requiring graphical mode

Conan-Kudo:
The lact daemon is usable in a command-line environment, since it can be interfaced with via the command line. As it's useful in headless setups, alter the systemd service unit to permit it.

### DIFF
--- a/res/lactd.service
+++ b/res/lactd.service
@@ -6,4 +6,4 @@ After=multi-user.target
 ExecStart=lact daemon
 
 [Install]
-WantedBy=graphical.target
+WantedBy=multi-user.target


res/lactd.service
```diff
@@ -1,9 +1,9 @@
 [Unit]
 Description=AMDGPU Control Daemon
-After=multi-user.target
+After=systemd-udevd.service
```
ilya-zlobintsev, line 3: I'm not sure if this is a good idea - there were already several issues caused by LACT being started too early in the bot process when the GPU was not fully initialized, and (at least on my system) the systemd-udev service starts several seconds earlier compared to the multi-user target.

Can you describe why the current service configuration does not work for you in headless setups? The `multi-user` target does not require a graphical session, and the `WantedBy` should only affect the graphical target it lists, not the service itself.

Conan-Kudo: The problem is that `WantedBy` declares what target it's part of when you use `systemctl enable lactd.service`. If you aren't booting into `graphical.target`, it won't get started.

ilya-zlobintsev: In that case, could both the `After` and `WantedBy` define `multi-user.target`? I've tested it with multi-user as the boot target and it seems to work like intended - the service is started only after multi-user is reached.

Conan-Kudo: Done.

PR #222, opened 2023-11-27
feat!: integrate libadwaita and its widgets and styles

GabMus:
Hello! I took the liberty of integrating libadwaita into LACT, but differently from #214 I didn't just add it to the project, I also redesigned most of the applicatiojn to use libadwaita's widgets and style where applicable.

I think that this way LACT does look better overall and it's possibly even a bit more usable, plus using libadwaita means one doesn't have to reinvent the wheel and can instead just use widgets that are already available in the library, following a consistent desing language.

Have a look at the screenshots below, let me know what you think, and thank you for the amazing work!

<details><summary>Screenshots</summary>

Screenshot from 2023-11-27 07-07-37
Screenshot from 2023-11-27 07-07-45
Screenshot from 2023-11-27 07-07-48
Screenshot from 2023-11-27 07-08-01
Screenshot from 2023-11-27 07-08-16
Screenshot from 2023-11-27 07-09-09
Screenshot from 2023-11-27 07-09-29
Screenshot from 2023-11-27 07-09-35
Screenshot from 2023-11-27 07-09-40
Screenshot from 2023-11-27 07-09-53
Screenshot from 2023-11-27 07-10-02
Screenshot from 2023-11-27 07-10-09

</details>


GabMus:
Ok, it seems ubuntu 22.04 has libadwaita version 1.1.7 and debian has version 1.2 while this patch requires >= 1.4.

What do you think about this? Do you see any way forward?

ilya-zlobintsev:
First of all, thank you for the contribution, it looks like you've put a lot of work into this PR.
I think overall it looks pretty good. I like most of the changes.

However, I have some concerns:
- The main issue that I have with libadwaita (as mentioned before in https://github.com/ilya-zlobintsev/LACT/pull/214#issuecomment-1826363673) is that it doesn't follow the system theme. Personally I do like how libadwaita looks, however, LACT is a desktop-agnostic tool, and using libadwaita would make it be inconsistent with other applications on non-gnome desktops (mainly KDE). There were already several people saying that they would prefer for LACT to keep using the system theme (for example in the linked PR). The current approach of using modern GTK but without libadwaita is a good middle ground IMO - it looks neutral in all desktops, and in gnome it can look pretty indistinguishable from libadwaita apps when using the adw-gtk theme.
- libadwaita is a relatively fast moving target. As you've already mentioned, some distros ship an older version of the library. Normally this is not an issue when apps are shipped as a flatpak or appimage, but  due to reasons described [here](https://github.com/ilya-zlobintsev/LACT/blob/master/pkg/README.md) LACT is currently fully reliant on distro packaging. This would mean that in practice when using libadwaita only a limited set of features supported on older versions will be available.
- Personally I'm not a big fan of the split view for navigation - there are most likely not going to be any new sections to navigate between, so most of the sidebar will remain empty. Tabs are a lot more efficient in terms of screen space usage in this scenario. While using a split view has the advantage of being automatically adaptable to different screen and window sizes, I don't think it's that important in LACT, as it's the kind of tool that in most cases will be used on a desktop with the window kept at its default size.

With that in mind, I don't think I would want libadwaita to be the default/only choice for the UI. But you've already put a lot of effort into it, and there are several options to go from here:
- Would it be possible to implement some of the UI changes you've made here without using libadwaita? I know it currently relies heavily on adwaita widgets, but it would be nice to keep some of the changes you've made here, like the setting categories in oc/thermals pages.
- I wouldn't mind having libadwaita be a build flag or a setting for people who do want to use it. This has the disadvantage of having to maintain slightly different variations of the UI, but if done properly the differences could be minimal (especially considering that most of LACT's logic is already in the daemon, and the GUI only sends commands to it)
- With the daemon in mind, it's possible to have a separate subproject/fork using libadwaita, but have it connect to an unchanged daemon.

ilya-zlobintsev:
> most apps on a desktop look inconsistent: steam, jetbrains ides, wine apps, any browser that's not falkon on plasma or epiphany on gnome, electron apps like element, and qt apps look just outright broken in anything but plasma

This is the unfortunate truth, but that doesn't mean that we need to contribute it. Also, for QT theming there are options to make it fit better in gnome, and there were initiatives to make it part of QT (though I'm not sure about their status)

>  they don't look any less inconsistent when they use the breeze theme on top

I disagree, at least they use the right colours and somewhat similar element styles

> it would be solved by https://github.com/ilya-zlobintsev/LACT/issues/224

Having a Flatpak would be nice, but I would still want there to be native packages that include the GUI for LTS distros

> a better approach would be to have parallel branches

I would be fine with this. If you (or someone else) are willing to maintain it, I can give permissions for a branch. There could also be a workflow to build packages and publish it on an alternative release.

Doing a build flag using cargo features to conditionally compile libadwaita versions of pages would be fine too if the differences are not that big (ideally it would just be different blueprint files using the same rust implementation)

Other approaches could work too. The main thing is that I want the plain GTK version to remain the "main" one, with libadwaita as an alternative.

GabMus:
> Also, for QT theming there are options to make it fit better in gnome

Off topic I guess :sweat_smile: but I've tried many different options myself, I found no way to make qt/kframework/kirigami apps look anything but broken outside of plasma

> I disagree, at least they use the right colours and somewhat similar element styles

Colors in libadwaita can be somewhat non-invasively changed with something like [Gradience](https://flathub.org/apps/com.github.GradienceTeam.Gradience), and in-app icons are still overwritten by the plasma theme without having to take any extra steps

> I would still want there to be native packages that include the GUI for LTS distros

Unfortunately, as of right now this looks like it's fundamentally incompatible with supporting older distros like debian stable and ubuntu lts.

> I would be fine with this. If you (or someone else) are willing to maintain it

I can try and keep this libadwaita version around, but I can give you no guarantees unfortunately, that's why I'd ideally want this merged rather than having it separate. A big refactor to the GUI would probably result in a huge mess of a merge conflict that's gonna be quite hard to solve

> different blueprint files using the same rust implementation

Eh, it's probably not gonna happen: there are fundamental differences between the two versions both in widget hierarchy and choice

lact-gui/Cargo.toml
```diff
@@ -15,7 +16,7 @@ gtk = { version = "0.7", package = "gtk4", features = ["v4_6", "blueprint"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 anyhow = "1.0"
-libadwaita = { version = "0.5.3", features = ["v1_4"] }
+libadwaita = { version = "0.5.3", features = ["v1_4"], optional = true }
```
ilya-zlobintsev, line 19: I think it would be convenient to alias `libadwaita` to `adw`, similar to how `gtk4` is renamed to just `gtk`. Makes it easier to type out full symbol paths

ilya-zlobintsev (commented):
Overall I'm glad that it looks feasible to support both libadwaita and non-libadwaita using feature flags - I think this solution will satisfy everyone

But I think it's worth splitting this into several pull requests - like setting up the infrastructure with the cargo features and package builds, adding some basic widgets, then adding more ui etc. Currently there are many somewhat separate things going on this PR - libadwaita, redesign of the non-libadwaita version, code refactoring/restructuring etc and it's a bit hard to keep track of everything at once

And another thing - after recently adding blueprint file support I've been migrating widgets to them one by one, so currently it's a mix of part blueprint and part in-code ui, and I think if there's a redesign of something then it's a good opportunity to use blueprint for the new/updated component where it makes sense (either for moderately big ui trees or for reusable components, for simple things it's sometimes not worth the required boilerplate)

lact-gui/src/app/root_stack/oc_page/power_cap_section.rs
```diff
@@ -62,6 +63,28 @@ mod imp {
...
+        #[property(get, set)]
+        pub default_value: RefCell<f64>,
+        #[property(get, set)]
+        pub value_text: RefCell<String>,
+
+        #[template_child]
+        pub adjustment: TemplateChild<OcAdjustment>,
+        #[template_child]
+        pub reset_button: TemplateChild<Button>,
+    }
```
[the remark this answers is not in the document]
GabMus: should be fixed

GabMus:
> Overall I'm glad that it looks feasible to support both libadwaita and non-libadwaita using feature flags - I think this solution will satisfy everyone

It ended up being more doable than I was expecting fortunately, but as you might expect the libadwaita version is a bit more polished, giving that there are actual widgets with proper implementations to support the new design language. In the long run I think it would be a good idea to just keep one version around, for sake of maintainability mostly. Plus I think that this might very well be a systemd-type situation, where there's some mystique around libadwaita and some people just hate on it since it's different and not what they're used to, even in terms of theming. I think once that's out of the way, and once distros catch up to newer versions of the library, it could be a good idea to drop the plain gtk version altogether, but that's ultimately your decision to make.

> But I think it's worth splitting this into several pull requests

Understandable, but I don't know how easy it would be or how much time it could take. To be completely honest while the diff is quite big I think that it's rather readable, mostly thanks to the excellent project structure. I'll see if I can get around to it.

> it's a good opportunity to use blueprint for the new/updated component where it makes sense

I can agree in principle, I'll see what I can do, but I can already tell you I'm not really too fond of the ergonomics of gtk templates in rust.

lact-gui/Cargo.toml
GabMus: should be fixed as well

GabMus:
Some new screenshots since stuff changed since the initial ones. Note that the icon in the about dialog is broken because the app isn't actually installed.

<details><summary>Plain GTK</summary>

Screenshot from 2023-12-02 13-31-58
Screenshot from 2023-12-02 13-32-08
Screenshot from 2023-12-02 13-32-13
Screenshot from 2023-12-02 13-32-26
Screenshot from 2023-12-02 13-32-36
Screenshot from 2023-12-02 13-32-40
Screenshot from 2023-12-02 13-32-45
Screenshot from 2023-12-02 13-32-54
Screenshot from 2023-12-02 13-33-38
Screenshot from 2023-12-02 13-33-43

</details>

<details><summary>Libadwaita</summary>

Screenshot from 2023-12-02 13-34-38
Screenshot from 2023-12-02 13-34-42

</details>

GabMus:
Any updates on this? How can we get this emrged?

ilya-zlobintsev:
I'd like to get this merged eventually, but first i need to finish up the packaging in #227. I also have some reservations about the design - I think it would be best if this PR only added the new libadwaita ui, and changes to the plain gkt one would be done separately
I'll take a look at it over the next few days

ilya-zlobintsev:
The packaging is done in #227 

As for the UI:

In the libadwaita version, I like how most of the UI looks. Some of the icons on the bottom navigation bar are missing on my system but that's probably an issue with my configuration.

For the OC page, there are some minor adjustments i'd like to have: i think it would make sense to group all of the clockspeed/voltage settings under one category, it doesn't really make sense to have a category just for the warning label and advanced button. 
There's a minor issue with the behaviour: it should only show settings available on the current gpu - for example, gpu voltage offset is only available on rx 5000+ series cards, while the absolute voltage value is only available on older gpus. Currently it hides rows if the max value is 0: [current implementation](https://github.com/ilya-zlobintsev/LACT/blob/b751f6ff91afe5a7c7b56e7b9b5ce6dd7d40e754/lact-gui/src/app/root_stack/oc_page/clocks_frame.rs#L387)

As for the plain GTK version, I'm not a big fan of all of the changes. I'd like to keep the headerbar tabs, and i think the current information lists design with less spacing between the items (and also less scrolling) looks better. 

Applicable to both versions: i'd probably want to keep the clocks/voltage adjustments are sliders, so there's a visual indication of the range that the values can have

GabMus:
> Some of the icons on the bottom navigation bar are missing on my system but that's probably an issue with my configuration.

Can confirm that they also seem missing on my end when running on my steam deck. It's likely some missing package in plasma, will see if there's anything that can be done about that.

> group all of the clockspeed/voltage settings under one category

Assuming you mean one listbox, this is what it looks like (sliders are coming back, too, next commit)

image

image

GabMus:
> There's a minor issue with the behaviour: it should only show settings available on the current gpu

Should be fixed

> I'd like to keep the headerbar tabs

Moved the stack switcher back to the headerbar, although this keeps the app a bit too wide IMO. I think we can come up with another solution, like maybe moving the stack switcher to a toolbar under the headerbar? Ultimately up to you.

> and i think the current information lists design with less spacing between the items (and also less scrolling) looks better.

I would disagree, having all that info smushed together like that decreases legibility without really adding much value: it's static data that you don't really need to keep an eye on, so having it spread out I think makes more sense.

A notable exception to this that I kept pretty much the same as the original is this section here:

image

Having this section more compact does add value: it's real time data that you may want to keep an eye on all at the same time, and that's why I didn't change it.

> Applicable to both versions: i'd probably want to keep the clocks/voltage adjustments are sliders, so there's a visual indication of the range that the values can have

Also done :)

GabMus:
The missing icon issue should also be fixed, the only thing that I'm not 100% sure about is that I had to install the thermometer icon for the temperatures tab. Unfortunately no built-in icon seems to be appropriate for temperatures, so the two options we have are:

a. install our own (that's what I'm doing in the last commit)
b. forego icons in the view switcher entirely

up to you

GabMus:
Any updates?

ilya-zlobintsev:
I'm sorry, I can't accept the PR as is. It should be split into smaller, incremental changes adding the new libadwaita UI.
I understand that you've put a lot of effort into it, but this is just too big of a change to do all at once.

nater1983:
@GabMus would u be willing to rebase this on the current version 0.5.1 I would like to add this application to Gnome for Slackware or GFS

PR #227, opened 2023-12-02
feat: add alternative headless and libadwaita builds

ilya-zlobintsev:


.github/workflows/build-packages.yaml
```diff
@@ -19,6 +19,10 @@ jobs:
         run: |
           curl -L -o /tmp/pkger.deb https://github.com/vv9k/pkger/releases/download/0.11.0/pkger-0.11.0-0.amd64.deb
           sudo apt -y install /tmp/pkger.deb
+          
+      - name: Set libadwaita target
+        if: contains(matrix.target-os, 'libadwaita')
+        run: sed -i 's/\&\& make/\&\& make build-release-libadwaita/g' pkg/recipes/lact/recipe.yml
```
GabMus, line 25: woulnd't it be better to keep the original recipe intact and just create another one for libadwaita?

ilya-zlobintsev: I'm trying to avoid the duplicate build steps between the recipes, ideally this would be done by passing some build flag when building the package, but unfortunately it doesn't seem possible

ilya-zlobintsev: I've ended up creating separate recipes after all so it's easier to manager

GabMus (approved):

GabMus:
LGTM :)

PR #241, opened 2024-01-09
chore: add keywords to .dektop file

FineFindus:
Adds a keyword section to the `.desktop` file. This allows for searching of the terms, instead of only the application name.


ilya-zlobintsev:
Thanks

PR #245, opened 2024-01-14
feat: close VulkanFeaturesWindow with shortcut

FineFindus:
Allows closing the VulkanFeaturesWindow with `ESC`. Since the search entry has the focus by default, and thus the window cannot receive the `ESC` shortcut, I added a workaround using the `stop_search` signal.


ilya-zlobintsev:
So am i understanding this correctly: by default, when the search entry is focused, the window gets closed with the `stop_search` signal, and if it's not in focus, it's handled by the ShortcutController on the window?

P.S. The CI failure is an intermittent error with the package builder, not related to the PR, should be fine after a restart

FineFindus:
Yes, that is correct.

PR #250, opened 2024-01-21
feat: display current_gfxclk if available

Umio-Yasuno:
Some AMD APUs/GPUs has the Average Clock and the Current Clock, the AMDGPU driver reports the Average Clock in `AMDGPU_INFO_SENSOR_GFX_SCLK` and hwmon.  
This PR gets the Current clock from `gpu_metrics` and displays it.

ref: https://gitlab.freedesktop.org/drm/amd/-/issues/2747

Screenshot_20240122_014943

### DIFF
--- a/lact-gui/src/app/root_stack/oc_page/gpu_stats_section.rs
+++ b/lact-gui/src/app/root_stack/oc_page/gpu_stats_section.rs
@@ -29,6 +29,7 @@ impl GpuStatsSection {
 
         let clockspeed = stats.clockspeed;
         self.set_core_clock(format_clockspeed(clockspeed.gpu_clockspeed));
+        self.set_current_core_clock(format_current_gfxclk(clockspeed.current_gfxclk));
         self.set_vram_clock(format_clockspeed(clockspeed.vram_clockspeed));
 
         let voltage = format!("{}V", stats.voltage.gpu.unwrap_or(0) as f64 / 1000f64);
@@ -89,6 +90,8 @@ mod imp {
         #[property(get, set)]
         core_clock: RefCell<String>,
         #[property(get, set)]
+        current_core_clock: RefCell<String>,
+        #[property(get, set)]
         vram_clock: RefCell<String>,
         #[property(get, set)]
         voltage: RefCell<String>,
@@ -129,4 +132,12 @@ mod imp {
 
 fn format_clockspeed(value: Option<u64>) -> String {
     format!("{}MHz", value.unwrap_or(0))
+}
+
+fn format_current_gfxclk(value: Option<u64>) -> String {
+    if let Some(v) = value {
+        format!("{v}MHz")
+    } else {
+        "N/A".to_string()
+    }
 }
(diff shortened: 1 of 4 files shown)


ilya-zlobintsev:
This seems similar to the previous change with the average/current power input readings. However, reading the drm repo thread, I don't think I entirely understand the behaviour: it looks like after the readings were changed to report the average clockspeed, the values became unstable, but shouldn't it be the other way around? It would make sense for the current reading to jump around, while the average should remain more or less the same. They also suggest to use `pp_dpm_sclk` to get the "current" readings, yet from what i understand that interface reports the target clockspeeds of power states, not the actual clockspeed.

Similar question regarding your screenshot: the current reading reports 800mhz, while the average is only 21mhz, that seems a bit weird. I would guess that the "current" value is the target clockspeed of the current power state, with the average being the actual (even if average) clockspeed.

Please tell me if i am misunderstanding how it works

Umio-Yasuno:
> They also suggest to use pp_dpm_sclk to get the "current" readings, yet from what i understand that interface reports the target clockspeeds of power states, not the actual clockspeed.

`pp_dpm_sclk 1:` displays the Current clock.

https://gitlab.freedesktop.org/agd5f/linux/-/blob/amd-staging-drm-next/drivers/gpu/drm/amd/pm/swsmu/smu11/sienna_cichlid_ppt.c#L1321-1343

> Similar question regarding your screenshot: the current reading reports 800mhz, while the average is only 21mhz, that seems a bit weird. I would guess that the "current" value is the target clockspeed of the current power state, with the average being the actual (even if average) clockspeed.

There is less documentation for `gpu_metrics`, but I think the same way.

Umio-Yasuno:
I think displaying the current target GFX clock can be helpful in in the OC setup.

lact-gui/ui/oc_page/gpu_stats_section.blp
```diff
@@ -40,6 +40,11 @@ template $GpuStatsSection: $PageSection {
                 value: bind template.core-clock;
             }
 
+            $InfoRow {
+                name: "GPU Core Clock (Current):";
```
ilya-zlobintsev, line 44: I think it would be better to name the values:
"GPU Core Clock" -> "GPU Core Clock (Average)"
"GPU Core Clock (Current)" -> "GPU Core Clock (Target)"
This way it's more clear on what the values mean

ilya-zlobintsev (changes requested):

lact-gui/src/app/root_stack/oc_page/gpu_stats_section.rs
```diff
@@ -130,3 +133,11 @@ mod imp {
...
     format!("{}MHz", value.unwrap_or(0))
 }
+
+fn format_current_gfxclk(value: Option<u64>) -> String {
+    if let Some(v) = value {
+        format!("{v}MHz")
+    } else {
+        "N/A".to_string()
+    }
+}
```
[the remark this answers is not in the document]
ilya-zlobintsev: I didn't see that this was changed to have extra logic with `u16`, with that it's fine as is

lact-gui/ui/oc_page/gpu_stats_section.blp
Umio-Yasuno: Ok, but the Average Clock and the Current Clock are clearly separated from Navi10.  
I think those values will be the same for pre-Navi10 APU/GPUs.

ilya-zlobintsev (approved):

PR #256, opened 2024-01-25
LACT OpenRC support

flakusha:
LACT daemon can work standalone on non-systemd distribution.

This is the example of service that is run in init environment similar to what is required for `sddm` / `gdm` which usually start the desktop session. It is possible to remove some of dependencies in case different configuration is required.

### DIFF
--- a/res/lact-daemon-openrc
+++ b/res/lact-daemon-openrc
@@ -7,7 +7,7 @@
 # # rc-service lact-daemon-openrc start
 
 supervisor=supervise-daemon
-command="/usr/bin/lact"
+command="lact"
 command_args="daemon"
 
 depend() {
--- a/res/lact-daemon-openrc
+++ b/res/lact-daemon-openrc
@@ -1,0 +1,23 @@
+#!/usr/bin/openrc-run
+
+# This service allows running LACT with Artix Linux OpenRC.
+# Place it under /etc/init.d and make it executable.
+# Run
+# # rc-update add lact-daemon-openrc 
+# # rc-service lact-daemon-openrc start
+
+supervisor=supervise-daemon
+command="/usr/bin/lact"
+command_args="daemon"
+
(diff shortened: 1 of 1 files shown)


res/lact-daemon-openrc
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/openrc-run
+
+# This service allows running LACT with Artix Linux OpenRC.
+# Place it under /etc/init.d and make it executable.
+# Run
+# # rc-update add lact-daemon-openrc 
+# # rc-service lact-daemon-openrc start
+
+supervisor=supervise-daemon
+command="/usr/bin/lact"
```
ilya-zlobintsev, line 10: Does this have to be an absolute path? The installation prefix can be different based on how LACT is is installed, e.g. if you just run `make install` this won't work as the binary will be under `/usr/local/bin`

flakusha: This is actually a good question.

This example is based on the [aur pkgbuild](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=lact) which is compatible with Artix Linux and *as many other aur build files* places new binaries in `/usr/bin`.

I can change to `/usr/local` which LACT uses originally, as there are other distros that use OpenRC and may default to work with `make`'s original path.

ilya-zlobintsev: Could it simply refer to `lact`, without the absolute path, so it uses wherever it's installed in PATH? That's what the [systemd service](https://github.com/ilya-zlobintsev/LACT/blob/master/res/lactd.service#L6) does.

flakusha: I have verified on my system that if LACT is already in $PATH, the OpenRC service works, code is updated.

ilya-zlobintsev:
Thanks

PR #270, opened 2024-02-22
fix: avoid selectable labels for dynamic info

Mershl:
Make InfoRow labels not selectable by default. Allow an override of the attribute in the template (for Software tab for example).

Motivation:
GTK will place the blinking cursor on the first selectable label of a view/tab. Placing the cursor on dynamic info like the current GPU Core Clock will show the cursor jumping around as the number of digits change a lot (best seen in the OC tab).

As the selection of a label is discarded on update of a label (which happens multiple times a second for a lot of the OC and Thermals values) the cursor acts as a visual distraction rather than a feature.

Demonstration of the issue:
Peek 2024-02-22 20-01

### DIFF
--- a/lact-gui/src/app/info_row.rs
+++ b/lact-gui/src/app/info_row.rs
@@ -37,6 +37,8 @@ mod imp {
         name: RefCell<String>,
         #[property(get, set)]
         value: RefCell<String>,
+        #[property(get, set)]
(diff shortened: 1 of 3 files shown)


ilya-zlobintsev (approved):
Makes sense, thanks

PR #279, opened 2024-03-03
RX 7900 Fixes and refactoring for error reporting

In-line:
Hi, Dear @ilya-zlobintsev thank you for amazing peace of open source software!

A few key points about this PR.

1. .context calls are added almost everywhere to make debugging issues easier
2. serde-error is added as dependency to produce nicer error backtraces
3. A weird bug fix for RX 7900 XTX

### DIFF
--- a/lact-daemon/src/server/gpu_controller/mod.rs
+++ b/lact-daemon/src/server/gpu_controller/mod.rs
@@ -651,9 +651,9 @@ impl GpuController {
             }
         } else if let Ok(hw_mon) = self.first_hw_mon() {
             if let Ok(default_cap) = hw_mon.get_power_cap_default() {
-                hw_mon
-                    .set_power_cap(default_cap)
-                    .with_context(|| format!("Failed to set power cap to default cap: {default_cap}"))?;
+                hw_mon.set_power_cap(default_cap).with_context(|| {
+                    format!("Failed to set power cap to default cap: {default_cap}")
+                })?;
             }
         }
 
--- a/lact-daemon/src/server/gpu_controller/mod.rs
+++ b/lact-daemon/src/server/gpu_controller/mod.rs
@@ -639,9 +639,13 @@ impl GpuController {
                 }
             }
 
-            hw_mon
-                .set_power_cap(cap)
-                .with_context(|| format!("Failed to set power cap: {cap}"))?;
+            // Due to possible driver bug, RX 7900 XTX really doesn't like when we set the same value again.
+            // But, also in general we want to avoid setting same value twice
+            if Ok(cap) != hw_mon.get_power_cap() {
+                hw_mon
+                    .set_power_cap(cap)
+                    .with_context(|| format!("Failed to set power cap: {cap}"))?;
+            }
 
             // Reapply old power level
             if let Some(level) = original_performance_level {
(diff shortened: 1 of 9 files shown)


In-line:
This is how errors look with this PR

image

lact-daemon/src/server/gpu_controller/mod.rs
```diff
@@ -639,7 +639,12 @@ impl GpuController {
                 }
             }
 
-            hw_mon.set_power_cap(cap)?;
+            // TODO: Due to possible driver bug, RX 7900 XTX really doesn't like when we set the same value again.
```
ilya-zlobintsev, line 642: I don't think it makes sense leaving this as a TODO - even if the issue with reapplying the value gets fixed on the driver side, there's no harm in keeping this check to avoiding unnecessarily writing the power cap.

Same in the `else` case

ilya-zlobintsev (changes requested):
Overall, I really like these changes, it should make tracking down issues in bug reports notably easier.

It looks like there's a test that needs to be updated: https://github.com/ilya-zlobintsev/LACT/actions/runs/8131682866/job/22222045972?pr=279

lact-daemon/src/server/gpu_controller/mod.rs
In-line: I removed TODO note, but general message is still there. I think in the future it makes sense to apply all changes incrementally.

lact-schema/src/tests.rs
```diff
@@ -34,12 +35,15 @@ fn controllers_response() {
...
+    let json_value = serde_json::to_value(&response).unwrap();
+    let parsed_response: Response<()> = serde_json::from_value(json_value).unwrap();
 
-    assert_eq!(serde_json::to_value(response).unwrap(), expected_response);
+    match (response, parsed_response) {
+        (Response::Error(first), Response::Error(second)) => {
+            assert_eq!(second.to_string(), first.to_string())
+        }
+        _ => unreachable!(),
+    };
```
ilya-zlobintsev, line 48: The goal of these tests is to check the actual json representation of responses, not just that serialization is working properly. So it should still have an `expected_response` json literal that's compared against the serialized form. 
I think here it would be especially useful to check the json representation of an error with a backtrace. If i'm understanding it right, it should be a nested error object.

In-line: @ilya-zlobintsev actual JSON representation is implementation defined in the serde-error crate. I think it doesn't matter what actual format is uses. You would need to update it after every serde-error dependency update.

ilya-zlobintsev:
The fedora package build error seems to be related to a missing signing key, you can ignore it.

lact-schema/src/tests.rs
In-line: What matters is that it produces same output after serialization/de-serialization.

In-line: I'm not sure how to implement that test properly without hard pinning the dependency

In-line: @ilya-zlobintsev do you have any suggestions?

ilya-zlobintsev: While the main purpose of the JSON API is communication between the GUI and the daemon, it's not an arbitrary internal API as it's possible to make third party scripts that interact with it. For this reason, I would like to avoid changing the JSON representation of requests/responses when possible. Of course this PR changes the error format, which is fine, but it would be best to avoid accidentally changing it again in the future.

In practice I don't think it's going to be an issue, serde-error didn't have any functional changes in [4 years](https://github.com/Ekleog/serde-error/commits/master/), but I would still like to have a test that defines what the error representation is.

In-line: @ilya-zlobintsev Okay, I pinned the dependency and updated test. I understand your point

ilya-zlobintsev (approved):

PR #280, opened 2024-03-03
fix: Fix statistics jumping around in GUI

In-line:
Hi, @ilya-zlobintsev a small fix, I hope it would be useful!

Before:
Before
After:
After


ilya-zlobintsev:
Didn't notice this during previous testing, thanks

PR #281, opened 2024-03-03
fix: Update stats from daemon only on OC page

In-line:
Why you ask? It was pretty annoying to comment out this loop, every time I was reading daemon debug logs.


ilya-zlobintsev:
Currently it's spammy when you enable debug logging, but maybe it's better to change the basic request/response logging to be `trace` instead of `debug`. Most functionality you're likely to be reading logs for is in the oc page anyway, so you'll probably have it open when changing the settings, and still have the logs spam (unless I'm misunderstanding what logs you're trying to read)

Also, the thermals page needs stats updates as well for temperature and fan speed monitoring

In-line:
Well, I think this PR isn't needed

PR #282, opened 2024-03-03
fix: Consistently format GPU statistics

In-line:
Hi, @ilya-zlobintsev  I understand, that this is purely subjective, but looking at Mhz and Voltage jumping around makes my eyes dizzy.

Before:
Before
After (+#280):
After


ilya-zlobintsev:
LGTM

PR #283, opened 2024-03-05
feat: Implement graph widget in OC tab

In-line:
graph

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -361,6 +361,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "byteorder"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
+
+[[package]]
 name = "bytes"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


In-line:
@ilya-zlobintsev I'm in the middle of refactoring and adding throttle visualization, but feel free to review WIP impl too

In-line:
New look & feel
plot

In-line:
@ilya-zlobintsev feel free to review

ilya-zlobintsev:
This is a very cool addition, but there are some considerations:
- The graph currently shows GPU usage and temperature on a single chart. This works fine since the values have a similar range, but in the future it could be useful to have graphs for clockspeed, voltage or power usage - those would have to be separate. Not yet sure how that should look
- IMO the OC page is getting kind of crowded, I think it would be best to hide the graph behind something like an expander by default, like it's done for power states. Alternatively, and considering point 1, there could be a dedicated popout window specifically for real-time graphs that the user can open on the side as needed. It could also contain more graphs in the future.
- The graph is currently drawn on a white background - ideally it would look like something that tries to integrate with the rest of the widgets by following the GTK theme, but I am not sure if that's feasible to implement. Though if the popup window is implemented this isn't as much of an issue, as there would be a separate window that's mostly white graphs, and the main window without them. Another solution could also be having dark/light modes for the graph, and selecting one based on system settings.
- Passing data to the plot via a json isn't the best. It should be possible to set it directly, but probably needs to be connected to the widget manually instead of relying on derived properties + bindings in the template
- I see that you had to downgrade gtk-rs for this. Probably can't do much about it until `plotters-cairo` gets updated

I'm also gonna need to read up some theory to understand how the cubic spline works :slightly_smiling_face: 

But overall this would be a very useful thing to have. Typically I use [amdgpu_top](https://github.com/Umio-Yasuno/amdgpu_top) for real-time monitoring, but it would be nice to have some historical data in LACT too, especially when you're over/underclocking and want to see the impact.

ilya-zlobintsev:
@In-line I've made another PR based on this one, implementing most of the things mentioned in the previous comment. Please take a look at it when you can: #301

PR #301, opened 2024-04-06
feat: implement historical charts window

ilya-zlobintsev:
Based on #283 and the ideas discussed in https://github.com/ilya-zlobintsev/LACT/pull/283#issuecomment-1979757300, except the graphs are in a separate window and show more information.
Screenshot_20240406_154226

The window can be opened either through the dropdown menu or a big button in the OC page.

The plotting logic is mostly unchanged since #283, except the value now gets modified directly instead of being sent over a GTK property as a JSON value.

The labels in the plot legend sometimes slightly clip through their container - I couldn't figure out how to fix this.


ilya-zlobintsev:
cc @In-line

In-line:
Thanks for taking your time to finish my work. I'm on a bit of hiatus now, as I have issues in finding time or motivation to contribute more. I will take a look at PR in few days. But please, be welcome to meld it in your opinioted way, without waiting for my opinion, as it may take a while.

Thanks for being amazing maintainer and shaping LACT into something I use daily

PR #389, opened 2024-10-20
Dev/threaded plot render

In-line:
* Rendering of plots moved to individual low-priority threads, which are able to drop frames
    * When resizing old texture is used when the new one isn't ready yet
    * On the first open a momentary black screen can appear (but it's better than freezing the window)
    * To further decrease CPU bottleneck cairo and plotters-cairo crates are compiled with opt-level 3
* This notably prevents LACT GUI unresponsiveness in the Graphs subwindow
   * On my system speedup for snapshot() function is 100x. From 4-5ms to 40-50us. 
* Plots look & feel further improved
    * Throttling graph jumping around is fixed by doing more data preprocessing instead of margin
    * Line thickness increased to 2
    * Supersampling fixed to correctly adjust to any sizes
    * Supersampling changed to 4x for better antialiasing
    * Legend is using filled rectangles now
* GUI benchmarks removed, sorry

better_plot


lact-gui/src/app/graphs_window/plot/imp.rs
```diff
@@ -57,17 +55,28 @@ impl WidgetImpl for Plot {
...
+            snapshot.append_texture(&texture, &bounds);
         }
     }
 }
 
-#[derive(Default)]
+#[derive(Default, Clone)]
 #[cfg_attr(feature = "bench", derive(Clone))]
 pub struct PlotData {
     line_series: BTreeMap<String, Vec<(i64, f64)>>,
```
ilya-zlobintsev, line 82: Perhaps it's better to wrap the data in an `Arc<Mutex>` or `Arc<RwLock>`, so that it can be passed to the render thread without cloning the whole BTreeMap and the strings inside it (which is an expensive operation)?

lact-gui/src/app/graphs_window/plot/render_thread.rs
```diff
@@ -0,0 +1,362 @@
...
+    pub height: u32,
+
+    pub supersample_factor: u32,
+}
+
+#[derive(Default)]
+struct RenderThreadState {
+    request_condition_variable: std::sync::Condvar,
+    last_texture: Mutex<Option<MemoryTexture>>,
+    current_request: Mutex<Option<Request>>,
```
ilya-zlobintsev, line 43: Is  `Condvar` + `current_request` behind a mutex somehow different from using a standard mpsc channel? It seems to be used exactly like a channel

lact-gui/src/app/graphs_window/plot/render_thread.rs
```diff
@@ -0,0 +1,362 @@
...
+}
+
+impl RenderThread {
+    pub fn new() -> Self {
+        let state = Arc::new(RenderThreadState::default());
+
+        let thread_handle = std::thread::Builder::new()
+            .name("Plot-Renderer".to_owned())
+            // Render thread is very unimportant, skipping frames and rendering slowly is ok
+            .spawn_with_priority(ThreadPriority::Min, {
```
ilya-zlobintsev, line 76: Just curious - what does the thread priority do on a system level? Does it set the `nice` value?

Previously I've used glib's `idle_add` for background work, but that runs on the main thread so it's not a good fit here.

lact-gui/src/app/graphs_window/plot/to_texture_ext.rs
```diff
@@ -0,0 +1,81 @@
...
+            unsafe {
+                let ptr = g_bytes_new_with_free_func(
+                    data.as_ptr() as *const c_void,
+                    (height * stride) as usize,
+                    Some(c_surface_destroy_notify),
+                    Box::into_raw(Box::new(surface_ref)) as *mut c_void,
+                );
+
+                bytes = Some(glib::Bytes::from_glib_full(ptr));
+            };
```
ilya-zlobintsev, line 48: I try to avoid manual memory management where possible, but this is fairly straightforward so it's probably ok

Cargo.toml
```diff
@@ -30,6 +30,12 @@ strip = "symbols"
 opt-level = "s"
 lto = true
 
+[profile.release.package.cairo-rs]
+opt-level = 3
+
+[profile.release.package.plotters-cairo]
+opt-level = 3
+
```
ilya-zlobintsev, line 38: Is it worth adding `plotters` itself to have opt-level 3 as well?

ilya-zlobintsev (changes requested):
Overall good changes, but I want to ask - was the plot rendering slow for you before?
On my system the UI remains responsive even in a debug build, the only thing that is laggy is resizing the charts window. 4-5ms sounds a bit high for the snapshot period, on my system it's 2.7ms according to the old benchmark even before adding opt-level to plotters.

ilya-zlobintsev:
The CI errors can be ignored, it's due to the fork not having a package signing secret.

lact-gui/src/app/graphs_window/plot/imp.rs
In-line: I would try that approach, it might simplify things a little bit. Would be better if all fields are wrapped around a single muted shared with render and GUI thread.

I just didn't check if gtk macros are supporting this use case.

Honestly, cost of rendering is so high, that I just don't see

lact-gui/src/app/graphs_window/plot/render_thread.rs
In-line: I think issue lies with my usage, as I like to fullscreen all the nice graphs we have.

And detaching it from GUI thread gives us room for higher supersampling. As currently this is 4x faster, because we render in parallel.

Cargo.toml
In-line: Wildcard with plotters and Cairo might do better here.

In-line:
> Overall good changes, but I want to ask - was the plot rendering slow for you before? On my system the UI remains responsive even in a debug build, the only thing that is laggy is resizing the charts window. 4-5ms sounds a bit high for the snapshot period, on my system it's 2.7ms according to the old benchmark even before adding opt-level to plotters.

I think issue lies with my usage, as I like to fullscreen all the nice graphs we have.

And detaching it from GUI thread gives us room for higher supersampling. As currently this is 4x faster, because we render in parallel.

lact-gui/src/app/graphs_window/plot/render_thread.rs
In-line: I think this is just a good practice to set explicit priority for this threads.

It didn't make difference for me personally, as I have high thread count CPU and almost never have CPU starvation for all of them. 

LACT before this PR was using 10% of single core for me (which is quite significant, considering I have AMD Ryzen 9 7950x CPU)

The idea is that graphs are not doing vital work and we are okay to run them at low frame rate. I wouldn't want all the fancy rendering to interfere much with someone's benchmarks.

As you understand, this is all educated guess work, but don't think it should have same priority as GUI thread.

In-line: It calls this syscall https://man7.org/linux/man-pages/man2/sched_setattr.2.html

ilya-zlobintsev: > LACT before this PR was using 10% of single core for me (which is quite significant, considering I have AMD Ryzen 9 7950x CPU)

This is weird, because on my slower 3700x it's <2% in a release build. But still it makes sense to draw this at a lower refresh rate, I agree.

lact-gui/src/app/graphs_window/plot/imp.rs
In-line: @ilya-zlobintsev On the second thought, that wouldn't work out, as we would still need to copy data, because in other case mutex lock would be held as long as frame is rendering, which isn't optimal at all.

Or we need to copy the data as soon as we acquire the mutex, which defeats the purpose.

Cloneing the data isn't that expensive to be fair in comparison to actual rendering

lact-gui/src/app/graphs_window/plot/render_thread.rs
ilya-zlobintsev: Tokio's `notify` channel type has this behaviour, but it doesn't have having blocking versions of the calls like the other channels (only async), so it won't work here. I suppose manual Condvar is fine

Cargo.toml
In-line: Wildcard pattern wasn't supported sadly

ilya-zlobintsev (changes requested):
On my system the text is way too big (at least at the default window size). Do you happen to use a high DPI screen with scaling? Maybe that's why it's different on your screenshots.

image

In-line:
Oh I see, I think we need to increase scaling bounds for the Window or decrease text size. I use Polonun tiling in KDE, it seems it was always resizing the Window, that's why I didn't notice.

-------- Original Message --------

In-line:
Default size: default size

Fullscreen: Fullscreen 1440p

ilya-zlobintsev (approved):
Tested on my system and it looks good

PR #390, opened 2024-10-22
Set codegen-units=1 to decrease binary size in release

In-line:
Before
```
❯ ls -lah target/release/lact         
-rwxr-xr-x 2 codemonkey codemonkey 6.9M Oct 22 18:45 target/release/lact
```

After
```
❯ ls -lah target/release/lact
-rwxr-xr-x 2 codemonkey codemonkey 6.4M Oct 22 18:46 target/release/lact
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ indexmap = { version = "2.5.0", features = ["serde"] }
 
 [profile.release]
 strip = "symbols"
+codegen-units = 1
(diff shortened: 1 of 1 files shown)


In-line:
There is no compilation time difference in my system

ilya-zlobintsev:
The release build compile times are dominated by LTO anyway, so this makes sense.

Issue #396, opened 2024-10-29
Fedora 41 Release


porkloin:
Fedora 41 is out now, is it possible to get an F41 tagged release?

ilya-zlobintsev:
I'll add it, but you can just use the older rpm for now as it will work fine in 99% of cases

porkloin:
Yeah sorry I forgot to add that when I opened the issue, but I was able to install the F40 rpm totally fine, works great on my ublue-based machine with `rpm-ostree install {lact-f40path}`

ilya-zlobintsev:
`dnf/rpm-ostree install lact` won't work regardless since the package isn't published in any repo, you always have to install it from a local file.

KyleGospo:
> `dnf/rpm-ostree install lact` won't work regardless since the package isn't published in any repo, you always have to install it from a local file.

Any interest in getting this into a copr repo? Would be huge for this purpose, and prevent the need to do updates like this since copr can fork for Fedora releases.

ilya-zlobintsev:
> Any interest in getting this into a copr repo?

If someone is willing to maintain it, then this would be great. It would be ideal if we could get LACT into Fedora repositories proper or into RPM fusion, but I'm not sure what the process for that is.

PR #403, opened 2024-11-07
Generate RPM Spec File for COPR

porkloin:
As discussed in [this issue about a fedora 41 release](https://github.com/ilya-zlobintsev/LACT/issues/396), this PR adds a bash script to generate a COPR compliant RPM spec from the existing pkger recipe files. This should (hopefully) make the burden of maintaining a COPR a lot less painful and not require you to manually edit three spec files on every release in addition to the recipe files.

Right now I temporarily have the generated spec file going into the release assets, but the next todo would be to make uploading it to COPR part of the release workflow. I wanted to get a draft version up before then just to get any thoughts/opinions if this is headed in the right direction or not since I've never contributed to the repo.

Lemme know if I need to make any adjustments!

### DIFF
--- a/.github/workflows/build-packages.yaml
+++ b/.github/workflows/build-packages.yaml
@@ -93,10 +93,16 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
+      - name: Install yq (for Fedora spec generation)
+        run: sudo apt-get install -y yq
+
       - name: Download artifacts
         uses: actions/download-artifact@v3
         with:
           path: downloaded-artifacts/
+
+      - name: Generate spec files for COPR submission
+        run: ./scripts/generate_spec.sh
 
       - name: Create release
         uses: ncipollo/release-action@v1.12.0
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ AppDir/
 appimage-build/
 *.stignore
 pkg/output
+pkg/fedora-spec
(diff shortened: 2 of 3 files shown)


ilya-zlobintsev:
I'm fine with doing this. My main concern with maintaining per-distro packages is that every time there's a new release each one has to be updated, and if there were any changes to the dependencies or build process since the last release they have to be adopted too. It already needs to be done for the Arch package and I didn't want to add any more.

Automatic generation from recipe files would resolve this, so once the package is created the maintenance should be minimal.

.github/workflows/build-packages.yaml
```diff
@@ -93,11 +93,17 @@ jobs:
...
+        run: sudo apt-get install -y yq
+
       - name: Download artifacts
         uses: actions/download-artifact@v3
         with:
           path: downloaded-artifacts/
 
+      - name: Generate spec files for COPR submission
+        run: ./scripts/generate_spec.sh
+
```
ilya-zlobintsev, line 106: Previously the `create-release` step was only actually used for updating the test release with new packages, and all the stable releases were created manually. I've changed this in https://github.com/ilya-zlobintsev/LACT/commit/23abed8ad6c718b5158dc637a83f5faec80a2607 so now there are two steps, one for all commits and one for stable tags. After this change the spec generation should go into the tag creation step, since I assume for now the COPR package will be just for stable versions.

porkloin:
Closing this one out in favor of #406 since I worked on it a bit more on that branch.

PR #406, opened 2024-11-09
Generate Spec Files for COPR on Release Publish

porkloin:
As discussed in [this issue](https://github.com/ilya-zlobintsev/LACT/issues/396) about a fedora 41 release, this PR adds a bash script to generate a COPR compliant RPM spec from the existing pkger recipe files. This should (hopefully) make the burden of maintaining a COPR a lot less painful and not require you to manually edit three spec files on every release in addition to the recipe files.

Notable changes or new code:
- `pkg/bin/generate_spec.sh` 
  - uses `yq` to extract a bunch of metadata about builds from your existing pkger file 
  - applies the extracter pkger values to a template COPR spec file. Notably, this extracts any build deps or deps that `fedora` or `fedora*` targets require
  - adds separate spec files for each recipe subdir in the pkg dir
  - Note: I haven't actually tested this with a real COPR build and am not super familiar with rust builds or tooling or the tooling of this repo, so you might want to look over the specs output in [this demo chore pr from my fork](https://github.com/porkloin/LACT/pull/4/files) 
  - I'm happy to update the template values if needed, just drop a review on this PR giving me any heads up on incorrect values
-  `.github/workflows/generate-spec.yaml` 
   - runs on any github release "publish" event
   - runs the generate spec bash script to generate new spec files
   - commits the spec files and opens a chore PR with the staged spec files
   - applies the `copr-update` label to the PR
- `.github/workflows/trigger-copr.yaml`
  - runs on any pr close
  - if the pr has the `copr-update` label, hit the webhook url configured in repo secrets at `COPR_WEBHOOK_URL`

This doesn't complete the entire integration to get a COPR build running, but it at least gets very close. AFAIK once the spec files are being reliably generated with accurate values for the build deps, etc, then we need to do some steps in COPR:

1. A COPR account and LACT repo (ex. ilya-zlobintsev/LACT) is created
2. Individual packages within that repo are created within COPR (ex. lact, lact-headless, etc) 
3. Each package is configured to point at this github repo in its SCM settings, and each package is configured to point at the correct file path for its spec file (i.e. `pkg/fedora-spec/lact`)
4. Set up a `COPR_WEBHOOK_URL` secret in your repo

@KyleGospo also mentioned that he may be able to do some of the above steps for us, because he has fedora pkger status which I think means he can config these to go into an official repo? idk enough about COPR or the internals of that to say.

Anyway, let me know if I need to update anything here!


ilya-zlobintsev:
The CI failure is unrelated and can be ignored

pkg/bin/generate_spec.sh
```diff
@@ -0,0 +1,99 @@
...
+
+%install
+rm -rf %{buildroot}
+make install DESTDIR=%{buildroot}
+
+%files
+%defattr(-,root,root,-)
+%license LICENSE
+%doc README.md
+/usr/bin/$RECIPE_NAME
```
ilya-zlobintsev, line 89: This needs to list all the installed files:
```
/usr/bin/$RECIPE_NAME
/usr/lib/systemd/system/${RECIPE_NAME}d.service
/usr/share/applications/io.github.$RECIPE_NAME-linux.desktop
/usr/share/icons/hicolor/scalable/apps/io.github.$RECIPE_NAME-linux.svg
/usr/share/pixmaps/io.github.$RECIPE_NAME-linux.png
```

pkg/bin/generate_spec.sh
```diff
@@ -0,0 +1,99 @@
...
+
+%changelog
+* $(date +"%a %b %d %Y") - $MAINTAINER - $GH_RELEASE_NAME - $GH_RELEASE_TAG
```
ilya-zlobintsev, line 92: It probably doesn't matter in github workflows, but I've had to change this to `LC_ALL=c date +"%a %b %d %Y"`, otherwise the changelog is generated with localized dates and RPM can't process them

pkg/bin/generate_spec.sh
```diff
@@ -0,0 +1,99 @@
...
+
+%prep
+%setup -q
+
+%build
+make %{?_smp_mflags}
+
+%install
+rm -rf %{buildroot}
+make install DESTDIR=%{buildroot}
```
ilya-zlobintsev, line 83: This has to be `make install PREFIX=/usr DESTDIR=%{buildroot}`, otherwise it installs into `/usr/local`

pkg/bin/generate_spec.sh
```diff
@@ -0,0 +1,99 @@
...
+BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
+ExclusiveArch:  x86_64
+BuildRequires:  $PKG_BUILD_DEPENDS
+Requires:       $PKG_DEPENDS
+
+%description
+$PKG_DESCRIPTION
+
+%prep
+%setup -q
```
ilya-zlobintsev, line 76: I've had to change this to `%setup -q -n LACT-%{version}` since the name of the source tarball doesn't match the folder inside it

ilya-zlobintsev (changes requested):
I think we should also generate spec files and commit them with the last released version (0.5.6). 

After the template changes and with version 0.5.6, the package builds successfully for me locally.

porkloin:
I think everything should be fixed up, thanks for the quick feedback! Please lmk if you think of anything else that needs to be updated. I still haven't heard from @KyleGospo about the COPR pkg maintainer stuff, but this at least gets everything in place. Once this is merged, if you wanna set up a COPR account in the interim and set up the repo and packages (one package for each of `lact-*` builds from recipes) and then set the github repo and spec file path for each package, at that point the webhook URL for the COPR repo should successfully scrape and build from source for the 0.5.6 release.

ilya-zlobintsev:
I've pushed a few small changes to your branch, and now copr builds successfully from your fork: https://copr.fedorainfracloud.org/coprs/ilyaz/LACT/
I will merge this and see how the automation works.

PR #423, opened 2024-11-24
Update io.github.lact-linux.desktop

Twig6943:
image


ilya-zlobintsev:
Thanks, forgot to update that description. It probably still says AMDGPU in a bunch of places

PR #425, opened 2024-11-26
Minor API documentation change, added example how to change profile.

Red007Master:
After addition of profiles API became much more relevant (for an average user).

On Windows people often have different profiles for different games and (I believe) Adrenaline allows you to change them automatically.

My (recently transitioned to Linux) friend wasn't able to figure out how to use API to change profiles.
So I decided to add it to docs.

I'm thinking about writing short guide how to set it up with Steam/Heroic if you are ok with it give me a green light.


ilya-zlobintsev:
Thanks, this is useful, though I'm not sure if it's worth making a full guide for Steam/Heroic, as once #424 is done it will cover this use case without having to do manual scripts.

PR #438, opened 2025-01-03
Dynamic loading of libdrm and libdrm_amdgpu

Umio-Yasuno:
`libdrm_amdgpu_sys 0.8.0` supports dynamic loading.  
Now `LACT` supports GPUs from multiple vendors, so dynamic loading would be better.

### DIFF
--- a/lact-daemon/src/server/handler.rs
+++ b/lact-daemon/src/server/handler.rs
@@ -939,7 +939,7 @@ fn load_controllers(
         match LibDrmAmdgpu::new() {
             Ok(libdrm_amdgpu) => {
                 info!("libdrm and libdrm_amdgpu initialized");
-                Some(Rc::new(libdrm_amdgpu))
+                Some(libdrm_amdgpu)
             }
             Err(err) => {
                 info!("AMDGPU support disabled, {err}");
--- a/lact-daemon/src/server/handler.rs
+++ b/lact-daemon/src/server/handler.rs
@@ -21,6 +21,7 @@ use lact_schema::{
     ClocksInfo, DeviceInfo, DeviceListEntry, DeviceStats, FanControlMode, FanOptions, PmfwOptions,
     PowerStates, ProfileRule, ProfileWatcherState, ProfilesInfo,
 };
+use libdrm_amdgpu_sys::LibDrmAmdgpu;
 use libflate::gzip;
 use nix::libc;
 use nvml_wrapper::{error::NvmlError, Nvml};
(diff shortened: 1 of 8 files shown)


ilya-zlobintsev:
Dynamic loading would make sense, though it is not a requirement as libdrm is something that is usually present on every system, even when using nvidia proprietary drivers.

However, I have one concern with this: what are the backwards compatibility guarantees of the libdrm API? Could the symbols change or be removed in an update? Currently this would lead to a link-time error, which is easier to check than a runtime one.

lact-daemon/src/server/handler.rs
```diff
@@ -932,6 +933,21 @@ fn load_controllers(
         }
     });
 
+    let libdrm_amdgpu = if sysfs_only {
+        None
+    } else {
+        match LibDrmAmdgpu::new() {
+            Ok(libdrm_amdgpu) => {
+                info!("libdrm and libdrm_amdgpu initialized");
+                Some(Rc::new(libdrm_amdgpu))
```
ilya-zlobintsev, line 942: I don't believe this needs to be wrapped in an `Rc`, as `LibDrmAmdgpu` is a cloneable struct with `Arc`s internally: https://github.com/Umio-Yasuno/libdrm-amdgpu-sys-rs/blob/main/lib.rs#L75

ilya-zlobintsev (changes requested):

lact-daemon/src/server/handler.rs
Umio-Yasuno: Done.  

https://github.com/ilya-zlobintsev/LACT/pull/438/commits/af7aecca4c0dd792e89afc8c536078afab77a072

Umio-Yasuno:
> Dynamic loading would make sense, though it is not a requirement as libdrm is something that is usually present on every system, even when using nvidia proprietary drivers.

For users who only use NVIDIA GPUs, there is less need to be concerned about `libdrm_amdgpu` by dynamic loading.  

> However, I have one concern with this: what are the backwards compatibility guarantees of the libdrm API? Could the symbols change or be removed in an update? Currently this would lead to a link-time error, which is easier to catch than a runtime one when a function symbol fails to load (or even worse, have the wrong signature).

Runtime errors can probably occur with dynamic linking, even if not dynamic loading.  
https://github.com/Umio-Yasuno/amdgpu_top/issues/17

ilya-zlobintsev:
I'll merge this as is since it turns out the mismatched header problem is already a possibility before this PR, but I would like to continue the discussion in https://github.com/Umio-Yasuno/libdrm-amdgpu-sys-rs/issues/9

Issue #447, opened 2025-01-15
[COPR] VERGEN_IDEMPOTENT_OUTPUT shown for lact commit id


Mershl:
### Checklist

- [X] The problem is not listed in the [hardware support matrix](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#hardware-support) as a known limitation. Examples of known driver-side issues are fans not turning on with a custom curve or inaccurate power limits on RDNA3.
- [X] I've included a [debug snapshot](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#reporting-issues) if the issue is related to hardware configuration

### Bug description

the rpm package from COPR (https://copr.fedorainfracloud.org/coprs/ilyaz/LACT) shows `VERGEN_IDEMPOTENT_OUTPUT` as commit id.

![image](https://github.com/user-attachments/assets/41502ead-a085-4359-b169-2fb607faf05d)

### System info

```markdown
lact 0.7.0
```

ilya-zlobintsev:
This is fixed in the `0.7.3-2` COPR build.

PR #462, opened 2025-02-02
pkg: manually set commit hash in fedora spec files

ilya-zlobintsev:
Fixes https://github.com/ilya-zlobintsev/LACT/issues/447


genbtc:
We still get hit by this same bug, in Gentoo (::Guru repo)
- likely need to apply the VERGEN changes you showed in the .spec file into the gentoo .ebuild if not main build system

ilya-zlobintsev:
> We still get hit by this same bug, in Gentoo (::Guru repo)
> 
>     * likely need to apply the VERGEN changes you showed in the .spec file into the gentoo .ebuild if not main build system

Yes, you just need to set VERGEN_GIT_SHA at build time.

PR #512, opened 2025-03-29
feat: configurable graphs window

ilya-zlobintsev:
Plots can now be added, removed, dragged around, and you can set which stats to show on each plot.
Note: the settings are currently not persistent 

image


ilya-zlobintsev:
@In-line I think you might be interested in this change

Issue #511, opened 2025-03-29
memory clocks can no longer go above 2.498 ghz


testc2n14:
### Checklist

- [x] The problem is not listed in the [hardware support matrix](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#hardware-support) as a known limitation. Examples of known driver-side issues are fans not turning on with a custom curve or inaccurate power limits on RDNA3.
- [x] I've included a [debug snapshot](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#reporting-issues) if the issue is related to hardware configuration
- [x] I've checked that the issue still happens in the latest [development build](https://github.com/ilya-zlobintsev/lact?tab=readme-ov-file#development-builds).

### Bug description

on my 7900 xtx i used to beable to have my clock go to 2700 but for some reason i can't get it to gove above 2.498, even if i set lact to 2501 (just 1 mhz higher then default) it will go up to that for a few seconds then go down to 1.544 ghz for a while then go back to 2.498. 

[LACT-v0.7.2-snapshot-20250328-201751.tar.gz](https://github.com/user-attachments/files/19514205/LACT-v0.7.2-snapshot-20250328-201751.tar.gz)

### System info

```markdown
- LACT version: 7.2 7.3
- GPU model: amd 7900 xtx merc 310
- Kernel version: 6.13.8
- Distribution: gentoo
```

Stellanora64:
I kinda have a similar problem with my 9070XT, what fixed it for me was lowering my monitor's refresh rate / resolution. (to be specific, I have my main monitor at 1440p at 144hz, and set my secondary monitor to 1080p at 60hz. Any higher would lock my memory clocks to around or just below stock).

This seems to be a driver / firmware issue, if that also fixes it for you.

testc2n14:
> I kinda have a similar problem with my 9070XT, what fixed it for me was lowering my monitor's refresh rate / resolution. (to be specific, I have my main monitor at 1440p at 144hz, and set my secondary monitor to 1080p at 60hz. Any higher would lock my memory clocks to around or just below stock).
> 
> This seems to be a driver / firmware issue, if that also fixes it for you (as memory overclocks work / apply fine after the display changes).

that didn't do it, the only 2 monitors conencted to my gpu is a 4k 120hz (can go up to 165 hz with dsc) and my 1080p 75 hz (running at 60 hz now)

testc2n14:
some more system info 
[...]

Issue #506, opened 2025-03-23
7800XT power cap not applying correctly


positroniumx:
### Checklist

- [x] The problem is not listed in the [hardware support matrix](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#hardware-support) as a known limitation. Examples of known driver-side issues are fans not turning on with a custom curve or inaccurate power limits on RDNA3.
- [x] I've included a [debug snapshot](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#reporting-issues) if the issue is related to hardware configuration
- [x] I've checked that the issue still happens in the latest [development build](https://github.com/ilya-zlobintsev/lact?tab=readme-ov-file#development-builds).

### Bug description

I have a strange issue, not sure if LACT or driver related.
Power slider goes to 280W but in testing power averages 250W. Tried various loads, but Kombustor testing shows this issue easily.
So, cat /sys/class/drm/card1/device/hwmon/hwmon5/power1_cap shows 280000000 but doesn't go past 250W average. Now, if I echo "280000000" > /sys/class/drm/card1/device/hwmon/hwmon5/power1_cap , basically the very same value, power consumption instantly jumps to 280W average, and clocks lower a bit.

![Image](https://github.com/user-attachments/assets/d8ca4681-ff9f-437a-b175-f39227606408)

In this photo can be seen where I issued the echo command, on the graph.
Also another issue which might be related to this, if I set constant fan speed and I run the previous echo command, the fan speed reverts to automatic.
cat /sys/class/drm/card1/device/hwmon/hwmon5/power1_cap shows the very same 280000000 value, so theoretically nothing changes as far as this values goes.
If I then drag the power slider in LACT, drag it lower then max it out again at 280 and press apply, the average power consumption again drops to around 250W. This can be seen in the following photo

![Image](https://github.com/user-attachments/assets/80af53f5-8a85-4947-b22b-d8d391057275)
So there is something strange going on which I don't really understand.

edit: forgot to mention that I have -190mV undervolt offset.
Tried with corectrl and when I apply -190mV offset in it it also drops to some 245W draw. power1_cap file still has 280000000 in it. So might be driver related?

### System info

```markdown
- LACT version: r677.6efbad8-1
- GPU model: 7800XT
- Kernel version: 6.13.7
- Distribution: Arch

GPU Model: AMD Radeon RX 7800 XT (0x1002:0x747E:0xC8)
[...]

Issue #500, opened 2025-03-18
VRAM Clockspeeds Are Not Applied Even Without Error Reported From LACT


anthonymonti2:
### Checklist

- [x] The problem is not listed in the [hardware support matrix](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#hardware-support) as a known limitation. Examples of known driver-side issues are fans not turning on with a custom curve or inaccurate power limits on RDNA3.
- [x] I've included a [debug snapshot](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#reporting-issues) if the issue is related to hardware configuration
- [x] I've checked that the issue still happens in the latest [development build](https://github.com/ilya-zlobintsev/lact?tab=readme-ov-file#development-builds).

### Bug description

After setting the clockspeeds and voltage the values using the GUI when I run a graphical program (furmak in this case) the memory does not hit the clock speed that was assigned.  I switched over from CoreCtrl as it doesn't support disabling zero RPM mode yet on the 7800xt. I noticed that when I set my overclock I get no reported error from LACT but when I run a game the memory clocks stay at stock. 
Here the situation. I moved my setting over to LACT (2660 MHz core, 2600 MHz vram). When I run a game with CoreCtrl I can see in amdgpu_pm_info the VRAM clock is correctly applied. Checking pp_od_clk_voltage shows 
```
OD_SCLK:
0: 500Mhz
1: 2660Mhz
OD_MCLK:
0: 97Mhz
1: 1300MHz
OD_VDDGFX_OFFSET:
-100mV
OD_RANGE:
SCLK:     500Mhz       5000Mhz
MCLK:      97Mhz       1500Mhz
VDDGFX_OFFSET:    -450mv          0mv
```
Then I closed CoreCtrl and killed its process to make sure its not running. Checking amdgpu_pm_info still shows 1300 MHz on the VRAM. As soon as I change any setting in LACT the VRAM frequency returns to stock (2436 MHz). Checking pp_od_clk_voltage shows the correct values like before but amdgpu_pm_info has reverted back to stock frequencies. My settings in LACT match what I was using in CoreCtrl. The performance profile stays 3D_FULL_SCREEN. I tried doing
```
echo "c" > /sys/class/drm/card1/device/pp_od_clk_voltage
```
just to see if maybe it was committing the changes but that didn't change anything. I don't know what CoreCtrl does differently when setting the values. Let me know if you need anything else from me thanks.

[...]

Issue #207, opened 2023-11-13
LACT sometimes completely disables power limit (6900XT)


adolfintel:
If the power limit is set to a value that's too low for the power state that the card is currently in, it seems that the driver completely disables the power limit, causing the card to have a spike in power consumption (I've seen it go as high as 420w up from a 293w TDP) until the card leaves that state and then it is applied correctly.

This can be easily replicated by setting the power limit while running any GPU intensive application that will keep the card at its highest power state (demonstrated in the video using stable diffusion) and can potentially lead to damage if running something like furmark which is not memory-bound.

OS: Arch Linux, Kernel 6.6.1, GPU: 6900XT

Here's a video of the problem:
[plbug.webm](https://github.com/ilya-zlobintsev/LACT/assets/16307689/bab92674-8010-46b4-b579-e54fd095e529)

ilya-zlobintsev:
As you've noted, this is the behaviour of the driver, which LACT has no direct control of. 
However, i think there can be a workaround: if the performance mode would be set to "lowest clocks" when a new power limit is applied, and then reverted back to whatever it was before (most likely "automatic"), then in theory this issue would be avoided.

Could you test if doing this manually avoids the problem on your system? Set the performance level to lowest together with the new power limit, apply, and then change it back to automatic. If it does help, then I'll implement it to be done automatically when setting the limit.

ilya-zlobintsev:
I don't think there's a good way to "suspend" the GPU.
Maybe there could be a wait interval until the clockspeed or power usage come down, though I'm not sure if it would be a good user experience. I'll see if i can reproduce this on my machine.

Kernel issues are generally reported here: https://gitlab.freedesktop.org/drm/amd

adolfintel:
Reported it: https://gitlab.freedesktop.org/drm/amd/-/issues/2992

ilya-zlobintsev:
@adolfintel if this is still relevant: could you try building the `power-limit-clockdown` branch and see if it solves the issue? I've added a workaround that waits until the GPU enters the lowest power state before applying the limit.

adolfintel:
@ilya-zlobintsev that works! Good job :)

PR #513, opened 2025-03-29
feat: updated amd settings apply order

ilya-zlobintsev:
Potential fix for #500 #506 #511

Generations tested:
- [x] Vega - Working
- [ ] RDNA1 - No hardware for testing
- [X] RDNA2 - No regressions
- [X] RDNA3 - Confirmed working by @positroniumx 
- [x] RDNA4 - No regressions

Note: the workaround for #207 is removed here, as it is not compatible with the new settings order, and the issue is closed as fixed in the driver.


positroniumx:
Everything works fine for me with 7800XT on arch. Now it goes 280W max.

ilya-zlobintsev:
@positroniumx could you also check if all the settings are still applied if you set performance mode to manual, and change the enabled p-states? That was also a problem in the past.

positroniumx:
Seems to work switching to manual, and selecting "power save" drops only the MEM clocks under load.

There's a strange issue present in both -git package and the testing package, with "Lowest Clocks" setting. When selecting it performance does drop, some 33% overall (benchmark performance), but GPU clocks shoot higher, it's strange. Like past set limit. I had it at 3050MHz limit and they go 3080-3100MHz. But performance drops and average power also drops some 10W, to 270W. This might be driver related.

There is another issue with both versions, something's wrong with Auto fan setting. It resets clocks on both versions, just on test version going back to fixed fan speed correctly reapplies everything else. But enabling auto fan speed resets clocks and power level, on both -git and test package.

So apart from the strange "Lowest clocks" behavior, and Automatic Fan Speed reseting clocks and everything else, seems fine.
I started reading documentation from here:
https://docs.kernel.org/gpu/amdgpu/thermal.html
Trying to understand various other options, I see there's a PCIe speed option as well, curious how it affects power levels for power save modes.
Got a complete list for all actions LACT applies to which files in /sys/class/drm/card*/device/ ?

ilya-zlobintsev:
>  something's wrong with Auto fan setting. It resets clocks on both versions

This was likely caused by the fan curve getting reset in the middle of applying other settings. Now the curve will get reset before all the other settings are written, this should fix the issue. New build will be available [here](https://github.com/ilya-zlobintsev/LACT/actions/runs/14154065413)

>  Got a complete list for all actions LACT applies to which files in /sys/class/drm/card*/device/ ?

The settings described in the [config documentation](https://github.com/ilya-zlobintsev/LACT/blob/master/docs/CONFIG.md) mostly map 1 to 1 to sysfs entries. In addition to the files listed in our earlier discussion in #506, LACT may also write to:
- `pp_power_profile_mode` - to set power profile modes, only used when "Performance level" is set to manual
- `pp_dpm_*` - to disable specific power states, also only available with manual performance level. The UI only exposes GPU and VRAM P-States (`pp_dpm_sclk` and `pp_dpm_mclk`), but it's possible to manually specify other p-state types in the config

LACT also writes to `pwm*` files inside of `hwmon` for fan control, but only on RDNA2 and older, as RDNA3/4 expose the new `gpu_od/fan_ctrl` files instead.

positroniumx:
I was able to reliably reset to auto by using 
```
echo "r" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "c" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
```
either
```
echo "r" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
```
I will try the new build you linked.
Any clues to what each bootup_default to window_3d options do differently? There's an issue when going back to Automatic Performance Level there's still the last Power Level Mode applied.
If I test Power_Saving on Manual, but switch from Manual to Automatic, there's still Power_Saving left on Power Level mode but greyed out, and its effects are still active, memory clocks don't boost they stay at Power_Saving mode level.
So I have to go back to Manual, switch back to more performant power level, then got back to Automatic.
I'll study the documentation some more to figure out how they all tie together. I'll also report back on your latest test build.

positroniumx:
Yeah there's also the occasional fail in switching between Automatic/Lowest/Highest/Manual and also between each Power Level modes, at least when under load. LACT interface completely freezes. Sometimes it recovers, sometimes it doesn't. Gotta kill the process then restart the interface (though not the service).

Not sure what that's about but not an issue atm, trying to figure out how the Power Level Mode/Performance Level tie together something seems off, especially with Lowest Clocks settings which boosts highest clocks on GPU while at the same time nerfing performance and lower 10W of average power draw in full load. Maybe related somehow.

ilya-zlobintsev:
> Any clues to what each bootup_default to window_3d options do differently?

They control power management heuristics. You can see the exact heuristic values that they have defined in the UI next to the name/in the `pp_power_profile_mode` file. We don't have much info about what these values do other than their names, but some people have reported in the past that a specific profile might work better in a given workload.

> There's an issue when going back to Automatic Performance Level there's still the last Power Level Mode applied.

That sounds weird, because the power mode file is not even writable when performance level is at auto, so theoretically the driver should not use it. Though it could be worked around to reset the mode specifically when going from manual performance level to non-manual.

positroniumx:
Ok tested latest build, all fine with auto fan control, now all controls on its panel work as well, they weren't before. Switching between auto/static doesn't affect the other settings. It's fine as it is in latest test build.

Coming back to performance level, Automatic doesn't seem to do much. Maybe it's not applying.

Here's Automatic/3D_fullscreen:

capture-2025-03-30_12-03-1743328158
capture-2025-03-30_12-03-1743328164

Highest Clocks/3D_fullscreen:

capture-2025-03-30_12-03-1743328244
capture-2025-03-30_12-03-1743328238
Seems clocks go a bit lower, power a bit higher.

Lowest Clocks/3D_fullscreen:

capture-2025-03-30_12-03-1743328362
capture-2025-03-30_12-03-1743328358
Clocks go highest for GPU, power drops to ~210W

Manual/Power Saving:
capture-2025-03-30_12-03-1743328448
capture-2025-03-30_12-03-1743328455
MEM clock goes down, and power drops some more to ~200W. GPU clocks still high af 3100MHz.

Automatic/Power Saving:
capture-2025-03-30_12-03-1743328546
capture-2025-03-30_12-03-1743328783

There's something weird about Automatic, it's like it doesn't apply or something. Or doesn't reset previous state. Also last photo you can see where LACT interface froze, luckily resumed and shows on graph for how long it was out.

positroniumx:
> That sounds weird, because the power mode file is not even writable when performance level is at auto, so theoretically the driver should not use it. Though it could be worked around to reset the mode specifically when going from manual performance level to non-manual.

That's what I was thinking. It may need a general reset thing. Something's off about Automatic coming back to it from other modes.
Also Lowest Clocks is clearly off, Highest clocks as well, there's come kind of inversion for GPU clock between them. In full load Highest Clocks drop GPU clock a bit, but Lowest Clocks max out the GPU clock.
Also mem shows at half what I set it at, is that some 2x thing that's not showing on graph?

positroniumx:
Looking back at issue #506 where I posted graph screenshots mem shows double the speed.

ilya-zlobintsev:
Yeah, the VRAM speed on the graph is a display only issue - the driver works with half rate values for GDDR6, LACT doubles them in the UI so it matches what people are used to in GPU specs/windows tools, and currently that isn't applied on the plot correctly, will fix it soon.

Either way, even with the weird performance level behaviour this already seems to be working way better than before. I will likely merge this once I'm able to confirm that it doesn't cause regressions on GPUs other than the 7000 series.

And do you have any relevant logs from when the UI freezes (either from the UI itself or the service)?

positroniumx:
I tried the gui from command line but nothing pops there when it hangs.
Service has something
```
WARN lact_daemon::server::handler: file "/sys/class/drm/card1/device/hwmon/hwmon5/power/autosuspend_delay_ms" exists, but could not be added to snapshot: Input/output error (os error 5)
```
```
lact[1876]: 2025-03-30T09:40:26.447534Z ERROR lact_daemon::server::handler: could not apply settings: Failed to set power cap: 280
lact[1876]: Caused by:
lact[1876]:     io error: Input/output error (os error 5)
```

Thing with performance level behavior is it might explain why some people on RDNA3 are having mem speed issues, Lower Clocks performance level drops them, and Power Saving power level, and if they go back to Automatic it might keep the low memory speed settings thus confusing them.
I'll have a play with performance and power levels in the meanwhile and report if I find something interesting.
So far It seems that going from Automatic to Lowest Clocks drops max power level, if then going on Manual it keeps the low power level, no matter which power level mode I select, which is also preserved when switching back to Automatic, but not in all cases, it's pretty confusing, I should start probing from command line and read the docs for that topic.

Ah yes, GUI hangs when cycling through these performance levels/power levels, it's fine with every other setting change.

positroniumx:
Figured out that LACT GUI doesn't correctly update the graphs for some reason.
In documentation I see there's a few extra profiles not just auto/high/low/manual. All of them are:
```
auto
low
high
manual
profile_standard
profile_min_sclk
profile_min_mclk
profile_peak
```
All seem to be working by themselves when activating with echo, don't have to be accompanied by the rest of commands. Setting power levels as well, at first glance.
Any chance of adding the missing ones as options?
Also anything else I could do regarding graph update? Confused me for a while since I followed it trying to figure out what's happening. Restarting GUI does seem to correctly reflect clocks/power levels, in graph.

ilya-zlobintsev:
> Figured out that LACT GUI doesn't correctly update the graphs for some reason (after issuing echo commands to change performance level).

Not all settings are polled regularly, performance level is only re-fetched on UI start or when you apply/reset settings 

>  Any chance of adding the missing ones as options?

Yes, they could be added 

> Also anything else I could do regarding graph update? 

Do you mean the halved VRAM clock? That should be fixed on latest master.

positroniumx:
>Do you mean the halved VRAM clock? That should be fixed on latest master.
Indeed this has been fixed.
Should I open a separate issue for "Lowest Clocks" issue? I have tried a few things and I am pretty sure there's something wrong with how it's applied.
I am monitoring pp_dpm__sclk looking at GPU clock.
If I set lowest clocks from LACT the GPU clocks go very high.
If I echo "low" > power_dpm_force_performance_level then GPU clock properly goes low.
When set from LACT the very same "low" value is in power_dpm_force_performance_level, so not sure what's going on. Seems I can send "low" by itself without other commands, and results are fine from command line.

positroniumx:
Here's one example:

```
#!/bin/sh
#echo "r" > /sys/class/drm/card1/device/pp_od_clk_voltage
echo "auto" > /sys/class/drm/card1/device/power_dpm_force_performance_level

#echo "profile_peak" > /sys/class/drm/card1/device/power_dpm_force_performance_level
echo "profile_min_sclk" > /sys/class/drm/card1/device/power_dpm_force_performance_level
#echo "profile_min_mclk" > /sys/class/drm/card1/device/power_dpm_force_performance_level
#echo "low" > /sys/class/drm/card1/device/power_dpm_force_performance_level
#echo "high" > /sys/class/drm/card1/device/power_dpm_force_performance_level
#echo "auto" > /sys/class/drm/card1/device/power_dpm_force_performance_level

echo "0 25 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "1 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "2 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "3 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "4 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve

echo "1" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_zero_rpm_enable
echo "c" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_zero_rpm_enable
echo "c" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
```

These two lines at beginning:
```
#echo "r" > /sys/class/drm/card1/device/pp_od_clk_voltage
echo "auto" > /sys/class/drm/card1/device/power_dpm_force_performance_level
```
Not sure yet if "r" is required, it resets to stock values before applying a profile.
Testing with "auto" seems like it does something interesting, when applying "profile_min_sclk" it inherits everything else from "auto" values, apart from difference, in this case min_sclk clocks.
The way this works out is like this. Say you had min min_mclk applied and apply min_sclk, it would inherit min_mclk clocks while also applying min_sclk. Resulting in low mem and gpu clocks.
But issuing an "auto" before, makes it so it inherits everything else from auto but min_sclk, thus you get an auto profile with min_sclk applied to it.
Which is more useful, instead they'd just compound, and makes for a total clusterfuck.
Inheriting auto everything makes for a sort of general state reset. A sort of echo "r" if you will.
I'm being verbose in the hope it helps you figure out what's happening, I'm still not sure but this seems to work this way.
I will test some more but curious what you think so far, makes any sense?

ilya-zlobintsev:
> Should I open a separate issue for "Lowest Clocks" issue? I have tried a few things and I am pretty sure there's something wrong with how it's applied.

Yes, please do so, i really appreciate the detailed investigation but it's a bit too much in this one issue :sweat_smile: 

> Theoretically clocks and undervolt and memory clocks should only ever be applied under "Manual" profile.

>  you can only ever undervolt with echo "vo -190" > /sys/class/drm/card1/device/pp_od_clk_voltage if you are on Manual profile in power_dpm_force_performance_level file, that shouldn't be possible in any other profiles, like Auto/High clock/Low lock.

Yes, that is stated in the documentation, however I don't believe this is entirely accurate - as you saw for yourself clock settings can work with the automatic performance level and have worked in the past with low/high clocks too on other generations. That is unlike the `pp_dpm_*`and `pp_power_profile_mode` files, which are only writable when performance level is manual, and will return an error otherwise.

Originally, performance level used to be applied at the start, but it was changed to be written after the clocks as this did work correctly together with custom settings, at least on RDNA2: #294 

The settings inheritance with the `profile` option sounds like the driver not cleaning up previous firmware state. It's a mess since the current state isn't reflected in the file but it shouldn't be a problem to reset the performance level to auto before applying any other settings.

> Power level shouldn't be written on "Low" profile, else clocks stay way higher.
But it should on "Auto/High/Profile_peak" options since that boosts performance, differently for each. Profile Peak makes for highest power but bit lower clocks, high makes for highest clocks but bit lower max power.

Not sure I understand what you mean later with the profile modes - should power cap not be applied at all when the profile is other than `auto/high/low`? That sounds weird, perhaps it should just be written before setting the mode?

Do you mean the "power level" setting in the UI? That corresponds to `pp_power_profile_mode`, it should only ever be writable with performance level at manual, LACT even forbids this:
```rust
            if let Some(mode_index) = config.power_profile_mode_index {
                if config.performance_level != Some(PerformanceLevel::Manual) {
                    return Err(anyhow!(
                        "Performance level has to be set to `manual` to use power profile modes"
                    ));
                }
```

Overall, logically and from your descriptions an order that's something like:
```
echo 'r' > pp_od_clk_voltage
echo 'auto' > power_dpm_force_performance_level # Reset
echo <user setting> > power_dpm_force_performance_level
echo <user settings> > pp_od_clk_voltage
<fan_ctrl settings>
echo '300000000' > hwmon/hwmon0/power1_cap
echo 'c'  > gpu_od/fan_ctrl/fan_curve and other fan_ctrl options
echo 'c' > pp_od_clk_voltage
```
Should work fine theoretically. I can make another PR to test this.

> I know this is a clusterfuck to program, I'm just telling you what I found.

Indeed, and the most annoying thing is that it's impossible to test without having access to the same or very similar GPU model, e.g. the behaviour I can observe on a 6900XT doesn't really correlate with what happens on RDNA3.

positroniumx:
>Not sure I understand what you mean later with the profile modes - should power cap not be applied at all when the profile is other than auto/high/low? That sounds weird, perhaps it should just be written before setting the mode?

If you apply a power cap value to Low performance level, the clocks don't properly go down, neither does power level.
When correctly set, Lowest Clocks performance level drops to 7W idle and both GPU and MEM clocks drop to ultra low values. Only without a power cap value.

For example like this:

```
#!/bin/sh
echo "auto" > /sys/class/drm/card1/device/power_dpm_force_performance_level
echo "low" > /sys/class/drm/card1/device/power_dpm_force_performance_level

echo "0 25 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "1 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "2 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "3 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "4 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve

echo "1" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_zero_rpm_enable
echo "c" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_zero_rpm_enable
echo "c" > /sys/class/drm/card1/device/pp_od_clk_voltage
echo "c" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
```

This correctly sets the Lowest Clocks Performance level. Resets to Auto then switches to Lowest Clocks.
This is valid for profile_min_sclk and profile_min_mclk which you are not using atm. With accompanying power cap value they fail to properly go down. They do something, but they do not go down to lowest sclk/mclk. Which they do without a power cap value applied at the same time.
The power cap value seems useful for Auto/High/Profile_peak
Profile_standard is kinda meh, gets GPU clocks to 1500MHz, nothing special, kind of mid level setting.

positroniumx:
Also
>echo 'r' > pp_od_clk_voltage

should be used only on manual, since it resets values related to manual profile. For resetting everything
```
echo "auto" > /sys/class/drm/card1/device/power_dpm_force_performance_level
```
Does the job before switching to another profile.

positroniumx:
Here's Low power profile properly activated:

```
!/bin/sh
echo "auto" > /sys/class/drm/card1/device/power_dpm_force_performance_level
echo "low" > /sys/class/drm/card1/device/power_dpm_force_performance_level

echo "0 25 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "1 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "2 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "3 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "4 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve

echo "1" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_zero_rpm_enable
echo "c" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_zero_rpm_enable
echo "c" > /sys/class/drm/card1/device/pp_od_clk_voltage
echo "c" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
```
capture-2025-03-31_15-03-1743422403

500MHz for GPU, 200MHz for MEM and 50W max TDP.

Here is with power cap applied:

```
#!/bin/sh
echo "280000000" > /sys/class/drm/card1/device/hwmon/hwmon5/power1_cap
echo "auto" > /sys/class/drm/card1/device/power_dpm_force_performance_level
echo "low" > /sys/class/drm/card1/device/power_dpm_force_performance_level

echo "0 25 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "1 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "2 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "3 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
echo "4 100 85" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve

echo "1" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_zero_rpm_enable
echo "c" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_zero_rpm_enable
echo "c" > /sys/class/drm/card1/device/pp_od_clk_voltage
echo "c" > /sys/class/drm/card1/device/gpu_od/fan_ctrl/fan_curve
```
capture-2025-03-31_15-03-1743422458

Last part of the graph, had to use max power profile so I can have a better contrast.
GPU clock stay at 2700MHz, MEM clock went to 1544MHz, TDP went to 170W.
That's a stark difference to proper Lowest Clocks profile levels. Nowhere near. I have tried applying the power cap in various places in the script, same thing, nothing gives as good results as omitting the power cap altogether, for whatever reason.

positroniumx:
Done the tests under full load. Amazing performance at 50W though. In a kombustor test I get 60% fps from full max power.

positroniumx:
Random confirmation for a 7600XT: https://bbs.archlinux.org/viewtopic.php?id=303301
Seems to be working on RDNA3.

PR #519, opened 2025-04-01
feat: Flatpak package

### LINKED ISSUE #517: Flatpak packaging, another attempt
It might be possible to package LACT as a Flatpak by shipping both the UI and daemon inside of the Flatpak and then using [flatpak-unsandbox](https://github.com/ryanabx/flatpak-unsandbox) to run the daemon on the host.
`flatpak-unsandbox` appears to work by:
- Converting all binary and library paths from inside of the Flatpak to the equivalent host path
- Injecting `LD_LIBRARY_PATHS` to use all Flatpak runtime paths, and running the binary via `ld-linux` from the Flatpak runtime but outside the sandbox

This should make it possible to run the daemon on the Flatpak host completely unsandboxed (and with root through `pkexec`) that still uses all the libraries shipped by Flatpak, and doesn't depend on anything but the kernel interfaces on the host side.

There would still be one notable limitation with this approach: no service management/autostart by default, as Flatpak does not support exporting services.
Perhaps there could be an option to install a service that runs the Flatpak unwrapper?

However, if this ends up working this approach would still be less problematic overall than all other universal packaging options: no large AppImage that depends on certain things from the host and is forced to be built on e.g. an old libc version, and no sandbox limitations that would come from running the daemon in any sort of container.

ilya-zlobintsev:
Closes #517 

This PR adds the following:
- Flatpak manifest to build LACT
- Metainfo XML file
- Special CLI command that generates a commandline that exports the path of the installed LACT flatpak and its runtime, so that the daemon can be executed on the host
- Bash helper script that sets up a systemd service if it is not detected when the Flatpak is launched
- Special `VULKANINFO_COMMAND` env variable, so that vulkaninfo can be overriden to a special `flatpak run` command in order for the daemon on the host to call back into the Flatpak environment 

All of this together lets us have an easy-to-setup LACT flatpak that's both universal and still has all the features from a daemon unconstrained by sandboxing.


flatpak/io.github.lact-linux.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - install -Dm755 flatpak/startup.sh /app/bin/startup.sh
+    build-options:
+      append-path: "/usr/lib/sdk/rust-stable/bin:/usr/lib/sdk/llvm20/bin"
+      prepend-ld-library-path: "/usr/lib/sdk/llvm20/lib"
+      env:
+        CARGO_HOME: "/run/build/lact/cargo"
+    modules:
+      - modules/yad.yaml
+      - modules/vulkan-tools.yaml
+      - modules/libdrm.yaml
```
tulilirockz, line 39: All of these should be under this same manifest, I've seen tons of flaptaks under the flathub org getting the "put everything in one place" treatment

flatpak/modules/libdrm.yaml
```diff
@@ -0,0 +1,9 @@
+name: libdrm
+buildsystem: meson
+builddir: true
+config-opts:
+  - -Dtests=false
+sources:
+  - type: git
+    url: https://gitlab.freedesktop.org/mesa/drm.git
+    tag: libdrm-2.4.124
```
tulilirockz, line 9: Never the tip of a branch. Reference a commit directly so this wont change between builds (even if the branch is EOLd)

flatpak/startup.sh
tulilirockz, line 1: I recommend having this being directly on a release instead of in the flatpak repo, they usually get mad at functionality being on the package instead of upstream

res/io.github.lact-linux.metainfo.xml
```diff
@@ -0,0 +1,32 @@
...
+  <summary>Linux GPU Control Application</summary>
+  <description>
+    <p>This application allows you to control your AMD, Nvidia or Intel GPU on a Linux system.</p>
+  </description>
+  <!--<screenshots>
+    <screenshot type="default">
+  <image>https://gitlab.gnome.org/World/design/contrast/raw/master/data/resources/screenshots/screenshot1.png</image>
+      <caption>Main Window</caption>
+    </screenshot>
+  </screenshots>-->
```
tulilirockz, line 16: You'll absolutely need this else your application wont show up on the flathub website (the releases too!).

lact-daemon/src/socket.rs
```diff
@@ -10,9 +10,9 @@ use tracing::{debug, info};
 pub fn get_socket_path() -> PathBuf {
     let uid = getuid();
     if uid.is_root() {
-        PathBuf::from_str("/var/run/lactd.sock").unwrap()
+        PathBuf::from_str("/run/lactd.sock").unwrap()
     } else {
-        PathBuf::from_str(&format!("/var/run/user/{uid}/lactd.sock")).unwrap()
+        PathBuf::from_str(&format!("/run/user/{uid}/lactd.sock")).unwrap()
```
tulilirockz, line 15: Maybe you could use `XDG_RUNTIME_DIR` for this socket?

.github/workflows/build-packages.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
tulilirockz, line 33: You should pin this, check out the `renovate` `best-practices` preset https://docs.renovatebot.com/upgrade-best-practices/

res/io.github.lact-linux.metainfo.xml
ilya-zlobintsev: Yeah this is just a placeholder for now

lact-daemon/src/socket.rs
ilya-zlobintsev: The user-specific path is a fallback that's only ever really used for development purposes, normally it's in `/run/lactd.sock`/`/var/run/lactd.sock`. I've had to change the path to `/run` in this PR because flatpak only allows exposing `/run` files and not `/var/run`, even if it's the same thing on most systems

flatpak/modules/libdrm.yaml
ilya-zlobintsev: Can I do a tag+commit together, to make sure the tag isn't force pushed? Having only a commit is not very descriptive in terms of what the version actually is

tulilirockz:
Separate, more general comment: I recommend adding `x-checker-data:` stuff to all your flatpak modules, it'll help you maintain it afterwards because you'll get automated PRs by the flathub bot. If you cant figure it out right now I can follow up this PR with the `x-checker-data` stuff :)

lact-daemon/src/socket.rs
tulilirockz: Alright! LGTM!

flatpak/startup.sh
ilya-zlobintsev: This is upstream, i'm only planning the manifest file to be present in the flathub repo, everything else will be here

flatpak/modules/libdrm.yaml
tulilirockz: I believe so! If you can actually do it I think its ok. Usually having a tarball is best tho

flatpak/startup.sh
tulilirockz: Perfect! Thats exactly what they want

flatpak/io.github.lact-linux.yaml
```diff
@@ -0,0 +1,39 @@
+app-id: io.github.lact-linux
```
tulilirockz, line 1: Appid is wrong, you need to specify the entire repo like:
```suggestion
app-id: io.github.ilya_zlobintsev.LACT
```

flatpak/io.github.lact-linux.yaml
```diff
@@ -0,0 +1,39 @@
...
+  - org.freedesktop.Sdk.Extension.rust-stable
+  - org.freedesktop.Sdk.Extension.llvm20
+command: startup.sh
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --talk-name=org.freedesktop.Flatpak
+  - --filesystem=/run/lactd.sock
```
tulilirockz, line 15: They might get mad at this, be prepared to have a nice explanation for it LOL

flatpak/io.github.lact-linux.yaml
```diff
@@ -0,0 +1,39 @@
...
+modules:
+  - name: lact
+    buildsystem: simple 
+    sources:
+      - generated-sources.json
+      - type: dir
+        path: "../"
+        skip: 
+          - "target"
+          - "target_ra"
```
tulilirockz, line 25: If this is just for debugging this is fine, but this will be a very big complaint if you dont fix this, see: https://github.com/flathub/io.github.doukutsu_rs.doukutsu-rs/blob/master/cargo-sources.json and https://github.com/flathub/io.github.doukutsu_rs.doukutsu-rs/blob/master/update-deps.sh

flatpak/io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,125 @@
...
+      prepend-ld-library-path: "/usr/lib/sdk/llvm20/lib"
+      env:
+        CARGO_HOME: "/run/build/lact/cargo"
+        PREFIX: "/app"
+    build-commands:
+      - cargo --offline fetch --manifest-path Cargo.toml
+      - cargo --offline build -p lact --features flatpak --release
+      - make install
+      - install -Dm755 flatpak/startup.sh /app/bin/startup.sh
+      - sed -i -E 's/lact gui/startup.sh/' /app/share/applications/io.github.ilya_zlobintsev.LACT.desktop
```
tulilirockz, line 42: I believe `.patch`es are preferred over `sed` if you can do that

flatpak/io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,125 @@
...
+        config-opts:
+          - -Dtests=false
+        sources:
+          - type: archive
+            url: https://gitlab.freedesktop.org/mesa/drm/-/archive/libdrm-2.4.124/drm-libdrm-2.4.124.tar.gz
+            sha256: 49c077f3938147e7c321fe89255eb189c1be68f6ed0aa36e29d38e3db0e84e08
+        x-checker-data:
+          type: anitya
+          project-id: 1596
+          url-template: https://gitlab.freedesktop.org/mesa/drm/-/archive/libdrm-$version/drm-libdrm-$version.tar.gz
```
tulilirockz, line 125: I think you need this directly on the `lact` module since its a dep for it, right? vulkan-tools should be fine not being on the lact module since its a separate tool

flatpak/io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,125 @@
...
+sdk-extensions:
+  - org.freedesktop.Sdk.Extension.rust-stable
+  - org.freedesktop.Sdk.Extension.llvm20
+command: startup.sh
+finish-args:
+  - --share=ipc
+  - --socket=fallback-x11
+  - --socket=wayland
+  - --device=dri
+  - --talk-name=org.freedesktop.Flatpak
```
tulilirockz, line 14: Just as a note here: We'll need to PR https://github.com/flathub-infra/flatpak-builder-lint/blob/master/flatpak_builder_lint/staticfiles/exceptions.json so this can be accepted on flathub

flatpak/io.github.ilya_zlobintsev.LACT.yaml
tulilirockz, line 125:
```suggestion
            x-checker-data:
              type: anitya
              project-id: 1596
              url-template: https://gitlab.freedesktop.org/mesa/drm/-/archive/libdrm-$version/drm-libdrm-$version.tar.gz
```

flatpak/io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,125 @@
...
+              - name: vulkan-headers
+                buildsystem: cmake-ninja
+                sources:
+                  - type: archive
+                    url: https://github.com/KhronosGroup/Vulkan-Headers/archive/v1.4.311/Vulkan-Headers-v1.4.311.tar.gz
+                    sha256: 7989c360b870bcc2de52b9fd626fd2ba7a212fe177ade3aa3a10b2bcb61f8689
+                x-checker-data:
+                  type: anitya
+                  project-id: 88835
+                  url-template: https://github.com/KhronosGroup/Vulkan-Headers/archive/v$version/Vulkan-Headers-v$version.tar.gz
```
tulilirockz, line 111:
```suggestion
                    x-checker-data:
                      type: anitya
                      project-id: 88835
                      url-template: https://github.com/KhronosGroup/Vulkan-Headers/archive/v$version/Vulkan-Headers-v$version.tar.gz
```

flatpak/io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,125 @@
...
+          - -DBUILD_ICD=OFF
+          - -DBUILD_VULKANINFO=ON
+        sources:
+          - type: archive
+            url: https://github.com/KhronosGroup/Vulkan-Tools/archive/v1.4.311/Vulkan-Tools-1.4.311.tar.gz
+            sha256: 7113bc0c746b45072e269fada0d684b4ae9de609c38d9e764b59793b14120a7b
+        x-checker-data:
+          type: anitya
+          project-id: 242111
+          url-template: https://github.com/KhronosGroup/Vulkan-Tools/archive/v$version/Vulkan-Tools-$version.tar.gz
```
tulilirockz, line 87:
```suggestion
            x-checker-data:
              type: anitya
              project-id: 242111
              url-template: https://github.com/KhronosGroup/Vulkan-Tools/archive/v$version/Vulkan-Tools-$version.tar.gz
```

flatpak/io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,125 @@
...
+            cleanup:
+              - '*'
+            sources:
+              - type: archive
+                url: https://launchpad.net/intltool/trunk/0.51.0/+download/intltool-0.51.0.tar.gz
+                sha256: 67c74d94196b153b774ab9f89b2fa6c6ba79352407037c8c14d5aeb334e959cd
+            x-checker-data:
+              type: anitya
+              project-id: 1385
+              url-template: https://launchpad.net/intltool/trunk/$version/+download/intltool-$version.tar.gz
```
tulilirockz, line 70:
```suggestion
                x-checker-data:
                  type: anitya
                  project-id: 1385
                  url-template: https://launchpad.net/intltool/trunk/$version/+download/intltool-$version.tar.gz
```

flatpak/io.github.ilya_zlobintsev.LACT.yaml
```diff
@@ -0,0 +1,125 @@
...
+            config-opts:
+              - -DVOLK_INSTALL=ON
+            sources:
+              - type: archive
+                url: https://github.com/zeux/volk/archive/vulkan-sdk-1.4.309.0.tar.gz
+                sha256: 1724924d8e3dccf0c508887edb79d56d9dd11b0738eab5a44e2fa95b8a9ebe1c
+            x-checker-data:
+              type: anitya
+              project-id: 370476
+              url-template: https://github.com/zeux/volk/archive/vulkan-sdk-$version.tar.gz
```
tulilirockz, line 100:
```suggestion
                x-checker-data:
                  type: anitya
                  project-id: 370476
                  url-template: https://github.com/zeux/volk/archive/vulkan-sdk-$version.tar.gz
```

flatpak/io.github.ilya_zlobintsev.LACT.yaml
ilya-zlobintsev: Moved vulkan-tools to top level

flatpak/io.github.ilya_zlobintsev.LACT.yaml
ilya-zlobintsev: I would like to keep everything in one file, i also considered making this a makefile step, but this probably makes more sense to be here since it's only used in the manifest

ilya-zlobintsev:
I'm gonna merge this now, this will make the .flatpak available through the test release. Later, after the next stable version release I'm gonna prepare a manifest for submission to Flathub and we can take a look at it then

PR #538, opened 2025-04-08
chore(lact-daemon): make nvidia bindgen feature as optional

eatradish:
to fix loongson3 and loongarch64 architecture build

### DIFF
--- a/lact-daemon/Cargo.toml
+++ b/lact-daemon/Cargo.toml
@@ -4,8 +4,9 @@ version = "0.7.3"
 edition = "2021"
 
 [features]
-default = []
+default = ["nvidia"]
 bench = ["dep:divan"]
+nvidia = []
 
 [dependencies]
 lact-schema = { path = "../lact-schema" }
--- a/lact-daemon/build.rs
+++ b/lact-daemon/build.rs
@@ -8,7 +8,10 @@ fn main() {
     println!("cargo::rerun-if-changed=include/");
 
     gen_intel_bindings();
+
+    #[cfg(feature = "nvidia")]
     gen_nvidia_bindings();
+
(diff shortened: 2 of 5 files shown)


ilya-zlobintsev:
The nvidia headers should be fully self-contained and not depend on anything on the current system, why does bindgen not work? Or do the headers use architecture-specific defines that fail on loongarch?

eatradish:
> The nvidia headers should be fully self-contained and not depend on anything on the current system, why does bindgen not work? Or do the headers use architecture-specific defines that fail on loongarch?

Indeed, your statement makes sense, the problem is most likely not nvidia's bindgen itself 

The reason for this is that others have previously reported that compiling LACT on loongarch64 gives an error similar to this:.

```
Compiling lact-daemon v0.7.3 (/root/LACT/lact-daemon)
error[E0308]: mismatched types
  --> lact-daemon/src/server/gpu_controller/nvidia/driver.rs:80:30
   |
80 |                 pAllocParms: ptr::from_mut(&mut alloc_params).cast::<c_void>(),
   |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u64`, found `*mut c_void`
   |
   = note:     expected type `u64`
           found raw pointer `*mut c_void`

error[E0308]: mismatched types
  --> lact-daemon/src/server/gpu_controller/nvidia/driver.rs:81:35
   |
81 |                 pRightsRequested: ptr::null_mut(),
   |                                   ^^^^^^^^^^^^^^^ expected `u64`, found `*mut _`
   |
   = note:     expected type `u64`
           found raw pointer `*mut _`
```

PR #582, opened 2025-05-18
feat: support profile rules with and/or conditions

### LINKED ISSUE #581: Add support for specifying multiple rules for a single profile (with AND/OR conditions)
Originally mentioned in https://github.com/ilya-zlobintsev/LACT/issues/574#issuecomment-2888544866, this would be a good addition to have

ilya-zlobintsev:
Closes #581 
image
image

### DIFF
--- a/lact-daemon/src/server/profiles.rs
+++ b/lact-daemon/src/server/profiles.rs
@@ -340,6 +340,12 @@ pub(crate) fn profile_rule_matches(state: &ProfileWatcherState, rule: &ProfileRu
                 }
             }
         }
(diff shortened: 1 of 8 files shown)


ilya-zlobintsev:
@Ph0rk0z you've previously mentioned not being able to set multiple processes in a comment, this is possible now.

Ph0rk0z:
Nice.. I will have to update. Now I can set clocks/undervolt by inference program and also unlocked clocks for idle (where it hopefully goes down from 30w). 

Is it possible to compare GPUs with graphs? Seems like it only allows for one.

PR #672, opened 2025-07-22
feat: Add PS4 GPU Liverpool&Gladius

ps4gentoo:
Playstation 4 GPU`s id`s (Phat & Pro)

### DIFF
--- a/lact-daemon/src/tests/data/pci.ids
+++ b/lact-daemon/src/tests/data/pci.ids
@@ -2324,6 +2324,11 @@
 	6658  Bonaire XTX [Radeon R7 260X/360]
 		1043 048f  R7260X-DC2OC-2GD5
 		1043 04d3  AMD Radeon R7 260X
+# Playstation 4
+	9920  Liverpool (Liverpool Playstation 4)
+	9922  Liverpool (Liverpool Playstation 4)
+	9923  Liverpool (Liverpool Playstation 4)
+	9924  Gladius (Gladius Playstation 4 PRO)
 # GV-R726XOC-1GD
 		1458 227b  Radeon R7 260X
 		148c 0907  Radeon R7 360


ilya-zlobintsev:
The pci.ids file is only used for running tests, and should be equivalent to the upstream pci db as the system one is actually used at runtime.
If you want to add a display name, it should be done in `res/device_ids.json`, though that currently only supports PCI subdevices.

BTW, could you send a debug snapshot from a PS4? I'm curious what is exposed there.

ps4gentoo:
Sorry was the wrong PR :D 
Yes here i have a DUMP  but yes we do have to make correct MESA patches
Vulkan and OpenGL is working but the Performance is not same like old mesa 22.x .... 

but i think overclocking does anyway not working so its just a monitoring tool for us because of SAMU and &ony secure... 

But nice Tool get me alot of infos about the GPU in the Dump, THANKS . 

This Dumps are from a PS4 Pro so the GLADIUS GPU 

[LACT-v0.8.0-snapshot-20250723-000653.tar.gz](https://github.com/user-attachments/files/21375468/LACT-v0.8.0-snapshot-20250723-000653.tar.gz)
[LACT-v0.8.0-snapshot-20250723-000718.tar.gz](https://github.com/user-attachments/files/21375469/LACT-v0.8.0-snapshot-20250723-000718.tar.gz)


The PS4 PRo GPU should be a gfx8 and not gfx7 (says CHATGPT like a Tonga or Polaris10) But i can not bring it run as gfx8 with the kernel only gfx7 .

but the GPU performance is not bad but have some Colorglitches because of wrong mesa patches, sometimes it doesnt have this Color Glitches and sometimes it have ... loooks like from the new Shader Compiler .... ACO 
https://youtu.be/CAzGNH3Qo4Q?t=3
just tested ....what does works....


and here a small video LACT on PS4 :) its on FEDORA i will compile at the weekend under gentoo if i have a little bit more time 
https://www.youtube.com/watch?v=Nixkrypxo4A


EDIT: but GPU and Mesa its to heavy for me to understand all things ... :-) and all this registers in the kernel for amdgpu in the cik.c File is alot work and heavy ... i have mutch respect for GPU Dev`s !! but yes they have the source and handbook  ;-) ...


would be nice if you can named the GPU`s in res/device_ids.json , have a nice Day! 
[LACTlog.txt](https://github.com/user-attachments/files/21375914/LACTlog.txt)

ilya-zlobintsev:
Added friendly names here: https://github.com/ilya-zlobintsev/LACT/commit/763b89e78f16881686088771b3267d0a4768fb8c

Based on the snapshots, it doesn't seem like anything is exposed in terms of control, but it's nice that at least the information is reported right.

PR #714, opened 2025-08-17
Add a service file for dinit

anoraktrend:
Dinit is yet another init system, which is used by chimera linux. lactd in the res folder is the service file for dinit. The file name can change if this is desired for clarity.


anoraktrend:
this doesn't work due to being unable to build on chimera linux for some reason i am unclear about.

ilya-zlobintsev:
Could this be shipped with a distro package rather than being included upstream? The openrc service was also contributed, but what I've found is that distro packages seem to either have to modify it, or ship their own one altogether.

anoraktrend:
Yeah probably. Lact doesn’t build on chimera Linux, but I’m too tired to try it again at the moment to make a proper issue about it. This particular file also doesn’t work with the flatpak.

PR #729, opened 2025-08-29
Add clippy check in CI

In-line:


.github/workflows/rust.yml
```diff
@@ -30,7 +30,22 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2
+      - name: Update repos
+        run: sudo apt update
+      - name: Install dependencies
+        run: sudo apt install -y libgtk-4-dev pkg-config libvulkan-dev libdrm-dev curl fuse3 libfuse-dev ocl-icd-libopencl1 ocl-icd-opencl-dev
+      - name: Install clippy
+        run: rustup component add clippy
+      - name: Run clippy
+        run: cargo clippy --verbose
```
ilya-zlobintsev, line 51: Could the check be done inside of the same job as the tests, in order to reuse the installed dependencies and build cache? Or is the intention to have it as a separate error from failed build/tests?

In-line: @ilya-zlobintsev intention is to have separate error, but you decide

In-line:
@ilya-zlobintsev would you be okay with me adding code coverage action? 

And I was also thinking about running cargo audit.

In-line:
@ilya-zlobintsev basically I want to add every possible check 😅

In-line:
I can do it in separate PR

ilya-zlobintsev:
I'm fine with adding extra checks, but please do it in a separate PR. Although I'm hoping there's a better way to set up the environment in each job than having to copy all the commands.
Something I've also wanted to add is build caching - you can take a look at that as well if you want, or I'll do it myself later.

.github/workflows/rust.yml
ilya-zlobintsev: The action is failing: https://github.com/ilya-zlobintsev/LACT/actions/runs/17332647025/workflow

In-line:
Sure, I like this project and your approach to it. Will fix ASAP and do improvements in separate PRs

PR #731, opened 2025-08-31
Add cache to CI

In-line:
Merge after https://github.com/ilya-zlobintsev/LACT/pull/729


ilya-zlobintsev (commented):
Could we use github cache to directly cache the target directory instead of sccache, for example via the [Rust cache action](https://github.com/Swatinem/rust-cache)?

As far as I understand standalone sccache will speed up compilation of individual crates, but it will still have to recompile even unchanged dependencies. Caching the whole target dir (as the cache action does) would avoid this, and should end up with faster builds overall

PR #736, opened 2025-09-03
feature: Optimize release with PGO

In-line:


ilya-zlobintsev:
I know this is a closed PR, but some thoughts on PGO:
- The benchmarks in LACT cover a very specific set of functionality where performance was important, but they are not really representative of everything the application does in real usage. So the PGO profile just from them might not be as effective as one from real application usage
- Collecting PGO profiles as part of the build action in makefile is definitely not a good idea, it would make building way too slow. If we were to have PGO (which I'm honestly not sure is worth it for LACT specifically) it would be some kind of a separate process to update the profile, and then reuse the profile for later builds

PR #740, opened 2025-09-06
feature: Add ASAN/TSAN to CI

In-line:


In-line:
@ilya-zlobintsev do you have any opinions here?

In-line:
@ilya-zlobintsev 

1) It's explicitly caused by each thread having their own `Handler`. They dlopen and dlclose the library concurrently in different test threads.
2) Thread sanitizer is probabilistic, it doesn't catch all issues 100% of the time. You probably need same environment as in CI. Locally it doesn't reproduces for me on Arch Linux.
3) It's arbitrary closest Rustc nightly that has all the necessary features, but isn't far from the one used in toolchain. We need pinned version to avoid instability. 
4) That's very good idea. I will reimplement PR based on that. 

Btw it's sad, that libdrm wrapper maintainer refused to acknowledge or fix the issue. They should have just marked the function as unsafe .

In-line:
@ilya-zlobintsev I found the issue in fuser Rust crate https://github.com/cberner/fuser/pull/390 it triggers both MSAN and LSAN

.github/workflows/rust.yml
```diff
@@ -97,3 +97,31 @@ jobs:
...
+        LSAN_OPTIONS: fast_unwind_on_malloc=0:max_leak_stack_depth=50:suppressions=../../../../lsan_suppressions.txt:print_suppressions=1
+        ASAN_OPTIONS: detect_leaks=1:detect_stack_use_after_return=1:check_initialization_order=1:strict_init_order=1
+      steps:
+        - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2
+        - uses: ./.github/actions/setup-environment
+        - name: Install Nightly Rust toolchain
+          uses: actions-rs/toolchain@v1
+          with:
+            profile: minimal
+            toolchain: nightly-2025-01-03
```
ilya-zlobintsev, line 119: It should use a current version of nightly as if right now (such as `nightly-2025-09-08`). I agree it should be pinned, but I don't think it has to match the MSRV actually used for builds, just for checks a new version is fine

lact-daemon/build.rs
```diff
@@ -22,6 +22,9 @@ fn gen_intel_bindings() {
         .header("include/intel.h")
         .parse_callbacks(Box::new(bindgen::CargoCallbacks::new()))
         .dynamic_library_name("IntelDrm")
+        // Don't generate bindings for functions and types related to file access
+        .allowlist_item("(DRM|drm).*")
+        .blocklist_item(".*(FILE|file).*")
```
ilya-zlobintsev, line 27: The generated functions don't get called, what's the harm in having them? It might be confusing in the future if some function that's present in the header is missing in the bindings

lact-daemon/src/server/gpu_controller.rs
```diff
@@ -131,16 +132,16 @@ pub struct PciSlotInfo {
...
 
 pub(crate) fn init_controller(
     path: PathBuf,
     pci_db: &pciid_parser::Database,
-    nvml: &LazyCell<Option<NvidiaLibs>>,
-    amd_drm: &LazyCell<Option<LibDrmAmdgpu>>,
-    intel_drm: &LazyCell<Option<Rc<IntelDrm>>>,
+    nvml: &'static Option<NvidiaLibs>,
+    amd_drm: &'static Option<LibDrmAmdgpu>,
+    intel_drm: &'static Option<IntelDrm>,
```
ilya-zlobintsev, line 144: This is incorrect - each respective library should only attempt to be loaded when a controller actually needs it. Right now it will print an error in logs how NVML couldn't be loaded even on a system that doesn't have an nvidia card.

But in general, I don't think it makes sense to even pass the libs as static references, you can just mark  the LazyLocks `pub(crate)` and access them from here or from controllers directly (though it's a good idea to still store the reference in the controller to avoid Option checks on every access)

lsan_suppressions.txt
```diff
@@ -0,0 +1,3 @@
+# Some leaks are detected from libfuse, they don't affect any real code
+# Fix is pending in https://github.com/cberner/fuser/pull/390
+leak:*libfuse.so*
```
ilya-zlobintsev, line 3: Please don't pollute the repo root dir, put this inline into the workflow if possible, or at least put it in a relevant subdir (maybe `lact-daemon/src/tests`)

ilya-zlobintsev (changes requested):
Side note: please avoid force pushing after a PR was already reviewed, it resets the review status and makes the notification links point to a non-existent diff

In-line:
Sure thing, will fix it later

In-line:
@ilya-zlobintsev Sanitizers don't work correctly on latest nightly you mentioned, I'm falling back to known working nightly version.

lact-daemon/src/server/gpu_controller.rs
```diff
@@ -131,16 +133,16 @@ pub struct PciSlotInfo {
...
 
 pub(crate) fn init_controller(
     path: PathBuf,
     pci_db: &pciid_parser::Database,
-    nvml: &LazyCell<Option<NvidiaLibs>>,
-    amd_drm: &LazyCell<Option<LibDrmAmdgpu>>,
-    intel_drm: &LazyCell<Option<Rc<IntelDrm>>>,
+    nvml: &'static LazyLock<Option<NvidiaLibs>>,
+    amd_drm: &'static LazyLock<Option<LibDrmAmdgpu>>,
+    intel_drm: &'static LazyLock<Option<IntelDrm>>,
```
ilya-zlobintsev, line 145: As I mentioned before already, you can directly reference the statics inside of this function (after making them crate-public), there is not much point to passing them as arguments now, it would make the function cleaner

lact-daemon/src/server/handler.rs
```diff
@@ -1055,72 +1058,81 @@ pub(crate) fn read_pci_db() -> Database {
...
+        error!("could not load Nvidia management library: {err}");
+        None
+    }
+});
 
-            Some((Rc::new(nvml), Rc::new(nvapi)))
+static AMD_DRM: LazyLock<Option<LibDrmAmdgpu>> = LazyLock::new(|| {
+    // SAFETY: We use global LazyLock to make sure it's safe.
+    #[allow(unused_unsafe)]
+    // author of LibDrmAmdgpu refuses acknowledge it's unsafe: https://github.com/Umio-Yasuno/libdrm-amdgpu-sys-rs/issues/12
```
ilya-zlobintsev, line 1103: Linking the libloading note would be useful: https://docs.rs/libloading/latest/libloading/struct.Library.html#thread-safety

And please avoid harassing the libdrm-amdgpu dev, their library has been extremely useful to LACT and they've contributed to the project in the past. It might be academically correct to mark the loader as unsafe, but it is indeed a situation where you are unlikely to get UB under real scenarios outside of parallel tests

Cargo.toml
```diff
@@ -55,3 +55,8 @@ strip = false
 debug = 1
 lto = "thin"
 codegen-units = 256
+
+# fuser contains memory safety issues which trigger leaks and UB inside libfuse.so
+# It's used only as dev dependency in the test code, so this is okay.
+[patch.crates-io]
+fuser = { git = "https://github.com/In-line/fuser.git", rev = "8daba15728da46d9299a2b23daa3aeb9954e114e" }
```
ilya-zlobintsev, line 62: Please don't point to your personal fork, i'd rather have the crate ignored by the sanitizer until the fix

ilya-zlobintsev: Your PR was merged, you can point to fuser master for now

PR #743, opened 2025-09-08
 feature: Implement CI check for feature powerset 

In-line:
Merge after https://github.com/ilya-zlobintsev/LACT/pull/742


ilya-zlobintsev:
Please avoid adding new checks after this PR unless there's a very good reason for it, there are already quite a lot

PR #756, opened 2025-09-20
feature: Speed up CI using nextest

In-line:


ilya-zlobintsev:
Can you explain what the benefit of this is? I've tried nextest locally and while it does seem nice and runs faster than the builtin cargo runner, the starting point with plain `cargo test` is ~1.1 seconds, which is not really anything noticeable in CI time... 

This adds yet another external dependency to the CI process. Even if it's a well-maintained high quality tool, it does mean that it will need to be kept up to date, including potentially adopting breaking changes (such as different parameters etc), or even the risk of it being deprecated or compromised. For these reasons I'm reluctant to add extra tools if the benefit is not very clear.

In-line:
@ilya-zlobintsev This is pretty much experimental. Maybe you're right. At this point there are not enough tests to justify it.

PR #757, opened 2025-09-20
feature: Use sccache for CI caching

In-line:


ilya-zlobintsev:
Do you have any before and after numbers for CI time? Comparing with another PR, I see 3m vs 2m for `build-test`, and ~9m to ~7m on `check-features-powerset`, is that what's expected?

In-line:
@ilya-zlobintsev pretty much, it would get better as project would get bigger.

PR #758, opened 2025-09-20
fix: set min width for thermals input

0xdeafbeef:
before
image
after
image


ilya-zlobintsev:
I believe the problem is originally caused by your GTK theme, as it looked fine with `Adwaita-dark` before for example:
image

However the width request looks innocent to have, so if it fixes the look for some people, I'm fine with having it.

PR #759, opened 2025-09-20
fix: use mono font so text doesn't shift when numbers are changed

0xdeafbeef:


ilya-zlobintsev:
Do you have any non-standard GTK fonts configuration? There should already be no shifting even without this.

0xdeafbeef:
> Do you have any non-standard GTK fonts configuration? There should already be no shifting even without this.

hm

```
gsettings get org.gnome.desktop.interface font-name
'Adwaita Sans 11'
```

image

ilya-zlobintsev:
I can reproduce the problem when using the adwaita font instead of noto sans, and `adwaita sans` is the default on recent gnome versions, so this is not a problem with your config.

lact-gui/src/app/info_row.rs
```diff
@@ -131,6 +131,7 @@ mod imp {
                         set_halign: gtk::Align::End,
                         set_use_markup: true,
                         set_ellipsize: pango::EllipsizeMode::End,
+                        add_css_class: "monospace",
```
ilya-zlobintsev, line 134: Please add this only for dynamic values (such as `gpu_stats_section`), it should not affect things like the hardware info page.

ilya-zlobintsev (changes requested):

0xdeafbeef:
@ilya-zlobintsev updated, wrapped only ints

ilya-zlobintsev (approved):
I like this, it's cleaner than putting a monospace class on the labels. Thanks

PR #761, opened 2025-09-20
feature: Build for Ubuntu 25.04

In-line:


ilya-zlobintsev:
I generally don't add non-LTS releases of Ubuntu to the package builds, as the version built for LTS works fine on them, since all the system libraries used are backwards compatible.

PR #784, opened 2025-10-20
(WIP) feat: JSON Schema generation

In-line:
1. JSON schema generation
2. Typed response/request protocol


ilya-zlobintsev:
Could you explain what the use case of JSON schema here might be? Generating type definitions for API clients in other languages?

In-line:
Yes, plus it's a documentation for API.

-------- Original Message --------

ilya-zlobintsev:
I have some other ideas about making the API easier to use, I'd like to try them out first, so this PR might sit unreviewed for a while

In-line:
Sure,let me know. When you would like this to get finished. There are few additional types that need custom spec providers

-------- Original Message --------

In-line:
@ilya-zlobintsev Hi, do you have any updates?

ilya-zlobintsev:
> @ilya-zlobintsev Hi, do you have any updates?

I had an idea to utilize the [varlink](https://varlink.org/) IPC protocol for LACT, as it is very similar to the current IPC implementation (JSON over unix sockets), but defined as a standard with the ability to define an interface spec that could easily generate client implementations.

This is still something I would like to do, but there were some questions left to figure out for a prototype, such as `varlink` vs `zlink` libraries and coexistence with the current IPC protocol.

Issue #719, opened 2025-08-19
Add more sensors to chart


pwlsoff:
I was thinking some extra sensors could be useful for some. There's more in gpu_metrics:
https://dri.freedesktop.org/docs/drm/gpu/amdgpu.html#gpu-metrics

Had a play with this script:
https://gist.github.com/leuc/e45f4dc64dc1db870e4bad1c436228bb

But had to change a few bits so most data is correctly shown:
[amdgpu_metrics.py](https://github.com/user-attachments/files/21857464/amdgpu_metrics.py)

Extra sensors that could be useful:

temperature_vrgfx
temperature_vrsoc
temperature_vrmem
voltage_soc
voltage_mem

ilya-zlobintsev:
Added in #798 alongside extra clockspeed types

PR #798, opened 2025-11-16
feat: add extra temperature and voltage readings on AMD

ilya-zlobintsev:
Closes https://github.com/ilya-zlobintsev/LACT/issues/719


bisspector:
lgtm

PR #851, opened 2026-01-08
fix: replace broken icons

makarov-roman:
found 2 more broken icons.
before(flatpak):
image
image
after:
image
image


ilya-zlobintsev (approved):

ilya-zlobintsev:
I never noticed this as the icons are present in KDE's default theme.

PR #852, opened 2026-01-08
feat: make oc page the main

makarov-roman:


ilya-zlobintsev:
I don't quite agree with this change. When you open the program for the first time, you're greeted with *some* set of values for adjustment, but it might not even be clear what GPU you have selected if it is a multi-GPU system. With the info page, it's at least obvious what you're looking at.
Considering the last tab is now saved with #842, is this really needed?

makarov-roman:
@ilya-zlobintsev   that's an interesting point. And it feels that regardless of the default page, we should show which gpu is selected somewhere on OC page (or in the header). What if we add gpu/driver to the GPU stats widget?
not saying we should copy the UI of afterburner. But just as a reference.
image

I started to do some minor changes on OC page, I'll soon send the first batch which will free a bit of space.

makarov-roman:
I would also argue that the restoration of the previous state makes sense where the state is important. e.g. IDE, browser.
But here I do think that having OC page as the entry point for the each session makes a lot of sense.

PR #853, opened 2026-01-08
feat: oc page move buttons to pageContent header

makarov-roman:
Hi, to avoid sending big PR I'm sending the first bits of changes.

with this change I want to make action buttons style more consistent, and free up some space which we'll need later.

before:
image
after-adw:
image

after-plain:
image

### DIFF
--- a/lact-gui/src/app/pages/oc_page/clocks_frame.rs
+++ b/lact-gui/src/app/pages/oc_page/clocks_frame.rs
@@ -16,7 +16,10 @@ use lact_schema::{
     request::{ClockspeedType, SetClocksCommand},
     ClocksTable, IntelClocksTable, NvidiaClockOffset, NvidiaClocksTable,
 };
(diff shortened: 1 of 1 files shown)


makarov-roman:
> For the reset buttons, I'm not sure. The power cap slider does not need more horizontal space

If you focus on power cap slider alone, then it look neat, but if you zoom out - you'll have 3 different approaches to buttons in just 3 widgets. That's why I wanted to standardize it. And I think it synergize well with the changes from #854
>  and I think it makes sense for the button to be inline. In the clocks configuration, either placement is probably fine, but when it's the bottom it is probably easier to find, since it's at the very end with nothing around it.

technically you can loose it under to apply toolbar, while the title is always visible. Also when you keep the same approach for UI elements, it makes interacting with it more intuitive. 
image

makarov-roman:
@ilya-zlobintsev I made the Reset button on OC widget red again. I'll address this again when I will work on the toolbar.
I'd suggest to focus on this MR before other(my) changes.

Issue #526, opened 2025-04-05
Drop libadwaita builds


ilya-zlobintsev:
Currently the `lact-libadwaita` packages link against libadwaita and initialize it, but don't actually use any libadwaita widgets. The only practical difference between `lact` and `lact-libadwaita` is the default theme.

You can achieve the same exact look as `lact-libadwaita` in the normal build by using [adw-gtk-theme](https://github.com/lassekongo83/adw-gtk3) (despite the name, it also provides a theme for GTK4). Considering this, it doesn't make a lot of sense to have a whole separate group of packages built just for a different default theme.

 I know that libadwaita is a controversial topic in the Linux community, and the stance for LACT is:
- LACT is a desktop-agnostic application that follows the system theme. For this reason, libadwaita is not the default option, as initializing libadwaita makes GTK ignore the system theme setting (though unlike commonly stated, it is still possible to style libadwaita apps with custom CSS, but it's a separate config from the overall GTK theme)
- libadwaita builds were provided for people who wish for LACT to integrate well with GNOME. For this purpose you can still use `adw-gtk-theme` and LACT will still look like a "native" libadwaita application.

PR #854, opened 2026-01-09
feat: update defaults for PageContent and InfoRow

makarov-roman:
Hi @ilya-zlobintsev ,
I want to get your opinion on the general approach before I invested too much into that. I dint split the changes because they makes less sense independently.

My goals:

- do not introduce UX changes or major structural changes (other than consolidation of some patterns)
- keep it as simple as possible, use generic gtk styles/components
- improve UX/UI consistency (e.g. use the same approach when we want to express something with the UI). 
- improve readability

I'll focus on OC page as this is the most important one. I do like how the changes affected the other pages, but there is some polishing required, and before that I want to get your confirmation on this one.

current UI as a reference:
image

updated-plain:
image

updated-adwaita: (show-historical-charts is hidded because it makes the layout a bit weird)
image

I'll list the changes so it's easier to discuss them:

#### 1. PageSection: 
add a container in the component to create padding and apply .card style
#### 2. InfoRow: 
use dim-label and caption css classes to improve readability of the component
this is especially visible on info pages (they are WIP, I haven't touched them except the component updates, I think it makes sense to use 2 columns here, and maybe consolidate software and hardware info pages, out of scope for now):
image
Screenshot From 2026-01-09 16-38-15
here on the second screen it's hard to read info because it's too far ahead.
also it let us use more space. Take a look at temps on gpu-stats widget. Before it was eclipsed, now we can fit all the data (at least for my gpu :smiley: )
#### 3. InfoRowLevel: 
add InfoRowLevel component. At first I moved vram usage indicator into InfoRow, but then realized that this component might be useful for other cases, it might be hard to read text when it's constantly updating. Applied it for stats which has already max values computed.
#### 4. GPUStats: 
I change the order of stats to pair relevant readings together. e.g. temp with fan speed, voltage and power usage, clocks together, gpu usage and vram usage. 
#### 5. GPUStats: 
I added GPU name based on your comment in #852 (Haven't figure out how to test dual gpu setup, I need to do some mock)

I would prefer if we can focus on the UI changes first, and once it's confirmed, we can focus on the code changes.

### DIFF
--- a/lact-gui/src/app/pages/software_page.rs
+++ b/lact-gui/src/app/pages/software_page.rs
@@ -144,7 +144,7 @@ impl relm4::SimpleComponent for SoftwarePage {
                                 set_column_spacing: 10,
                                 set_homogeneous: true,
                                 set_min_children_per_line: 2,
-                                set_max_children_per_line: 2,
+                                set_max_children_per_line: 4,
                                 set_selection_mode: gtk::SelectionMode::None,
 
                                 append_child = &InfoRow {
--- a/lact-gui/src/app/pages/oc_page/gpu_stats_section.rs
+++ b/lact-gui/src/app/pages/oc_page/gpu_stats_section.rs
@@ -42,7 +42,7 @@ impl relm4::SimpleComponent for GpuStatsSection {
                 set_column_spacing: 10,
                 set_homogeneous: true,
                 set_min_children_per_line: 2,
-                set_max_children_per_line: 2,
+                set_max_children_per_line: 4,
                 set_selection_mode: gtk::SelectionMode::None,
 
                 append = &InfoRow {
--- a/lact-gui/src/app/pages/thermals_page.rs
+++ b/lact-gui/src/app/pages/thermals_page.rs
@@ -121,7 +121,7 @@ impl relm4::Component for ThermalsPage {
                         set_column_spacing: 10,
                         set_homogeneous: true,
                         set_min_children_per_line: 2,
-                        set_max_children_per_line: 2,
+                        set_max_children_per_line: 4,
                         set_selection_mode: gtk::SelectionMode::None,
 
                         append_child = &InfoRow {


ilya-zlobintsev:
Please do remember that AMD cards expose more data, previously on the information page everything fit with plenty of space left, and now it doesn't:

image
image

As you mentioned, it might make sense to show the data in 2 columns.

For the OC page you can test how all the stats look by using test data with `_LACT_DRM_SYSFS_PATH=./lact-daemon/src/tests/data/amd/rx9070 cargo r` (daemon has to be stopped for the data to be used)

Overall I think the OC page looks better with this, but there is some information missing, such as total VRAM size. And I'm not sure about showing the GPU name there, though I understand this is from the discussion on #852

Also the number of columns appears to scale with the text size, on my laptop iGPU it's showing 3 columns even at minimum window width:
image

makarov-roman:
image
@ilya-zlobintsev do I need to install anything? this is on current master

ilya-zlobintsev:
> image @ilya-zlobintsev do I need to install anything? this is on current master

For the info page most data can only be fetched from a real connected GPU, so mock data doesn't work. But the stats on the OC page should be displayed.

makarov-roman:
@ilya-zlobintsev I think I won't add any other improvements to this PR. 

~ ~the only thing which I wanted to fix in scope of this PR gave up in the process is this
This stack component stretches by the content of the biggest stack element. I tried to use the same approach as for info_row with set_visible, but it's not that straightforward. Maybe you have an idea?~ ~  fixed

can you provide the feedback for UI and code?

makarov-roman:
> Overall I think the OC page looks better with this, but there is some information missing, such as total VRAM size. And I'm not sure about showing the GPU name there, though I understand this is from the discussion on https://github.com/ilya-zlobintsev/LACT/pull/852

I removed some data on purpose. My idea is that OC page should be tailored for the OC process (I want to suggest to remove more data which is not relevant to `apply-oc -> check how it behaves(temps, clocks, pow) -> apply oc` loop, but later in another PR).  

max vram is static value, and not relevant to OC, therefore I decided to remove it to make the text more readable. (users can still see it on HWINFO page)

lact-gui/src/app/pages/software_page.rs
```diff
@@ -47,91 +52,82 @@ impl relm4::SimpleComponent for SoftwarePage {
...
                 PageSection::new(&fl!(I18N, "system-section")) {
-                    append = &InfoRow::new_selectable(&fl!(I18N, "lact-daemon"), &daemon_version),
-                    append = &InfoRow::new_selectable(&fl!(I18N, "lact-gui"), &gui_version),
-                    append = &InfoRow::new_selectable(&fl!(I18N, "kernel-version"), &system_info.kernel_version),
+                    append_child = &gtk::FlowBox {
+                        set_orientation: gtk::Orientation::Horizontal,
+                        set_column_spacing: 10,
+                        set_homogeneous: true,
+                        set_min_children_per_line: 2,
+                        set_max_children_per_line: 2,
```
ilya-zlobintsev: Does it make sense to force only two columns? I think it's okay to show more when the window is resized horizontally. 
This applies to all of the FlowBoxes, it just needs to have appropriate size limits (e.g. so the default window size always ends up with 2 columns)

lact-gui/src/app/pages/software_page.rs
```diff
@@ -47,91 +52,82 @@ impl relm4::SimpleComponent for SoftwarePage {
...
-                                append = &gtk::Label {
-                                    set_halign: gtk::Align::Start,
-                                    set_hexpand: true,
-                                    set_label: &format!("{}:", fl!(I18N, "extensions")),
                                 },
 
-                                append = &gtk::Button {
+                                append_child = &InfoRow {
+                                    set_value: fl!(I18N, "extensions"),
+                                    set_icon: "go-next-symbolic".to_string(),
```
ilya-zlobintsev: I think with the current style of InfoRow + icon, it's not very obvious that it's clickable. At best, it may look like the icon is clickable, when in reality it's the whole row. It should have an outline, different colour or some other way to show that it's actually a button.

lact-gui/src/app/pages/software_page.rs
makarov-roman: @ilya-zlobintsev for GPU stats I prefer to have just 2, because they have reasonable structure. For info/sw pages - I think no need for the max limit. will change

lact-gui/src/app/pages/software_page.rs
ilya-zlobintsev: The plain version is the main one, packages for libadwaita are not even provided anymore, see #526.

Generally speaking, if things look okay in the base version, they will also look good with libadwaita. I'm still keeping the build option around because there are packages made for it (at least in Arch) and it does ultimately look better than the plain version by itself, at the cost of not following the system theme. The linked issue has more reasoning about the choice.

lact-gui/src/app/info_row.rs
```diff
@@ -23,6 +25,67 @@ impl InfoRow {
...
+
+    // idk why it complains about these not being used, but they are
+    #[allow(dead_code)]
+    fn set_name(&self, name: String);
+    #[allow(dead_code)]
+    fn set_value(&self, value: String);
+    #[allow(dead_code)]
+    fn set_icon(&self, icon: String);
+    #[allow(dead_code)]
+    fn set_level_value(&self, level_value: f64);
```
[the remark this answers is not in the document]
ilya-zlobintsev: You kept the dead code attribute on these functions, when they can be removed along with their implementations and everything still compiles and works.

Generally the overwhelming majority of rust's warnings/lints are correct, it's very rare that you have to exclude them.

lact-gui/src/app/pages/software_page.rs
makarov-roman: @ilya-zlobintsev the problem what I see it - there is no width limits in the container. It looks already a bit weird on qhd monitor, but if somebody maximize the window on these ultrawides -  the layout would be completely broken.
I added a static width on the main Box, and I do like how it behaves. Can you check?

lact-gui/src/app.rs
```diff
@@ -179,6 +180,12 @@ impl AsyncComponent for AppModel {
         root: Self::Root,
         sender: AsyncComponentSender<Self>,
     ) -> AsyncComponentParts<Self> {
+        if !cfg!(feature = "adw") {
+            if let Err(err) = relm4::set_global_css_from_file("lact-gui/res/style.css") {
+                tracing::warn!("Failed to load global CSS from file: {err}");
+            }
+        }
```
ilya-zlobintsev: This won't work anywhere other than when running from the repo. The CSS should be included in the binary (likely with `include_str!`) and loaded from that string.

lact-gui/src/app/pages/software_page.rs
ilya-zlobintsev: > I switched to 2/4 columns on soft/hardware pages + gpu_stats

It appears to still be max 2 columns in the OC page stats, it's inconsistent between pages due to this currently

ilya-zlobintsev: Also. you missed the OpenCL info frame, it's inconsistent within the software page as well

PR #861, opened 2026-01-18
feat: update apply toolbar and move scrollwindow to apps.rs

makarov-roman:
Hi,
only cosmetic changes here. I liked the idea from #222 to move apply/revert to header and use bottom toolbar for nav, but I don't want to do any structural changes until there is a solution regarding adw.
1) use ActionBar for applyRevealer widget.
2) move ScrolledWindow to app.rs and fix margins around it
3) change buttons, change their order (usually, happy path button is on the right)

### DIFF
--- a/lact-gui/src/app.rs
+++ b/lact-gui/src/app.rs
@@ -135,6 +135,9 @@ impl AsyncComponent for AppModel {
 
                         #[name = "root_stack"]
                         gtk::Stack {
+                            set_vexpand: false,
+                            set_vhomogeneous: false,
+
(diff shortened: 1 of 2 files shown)


makarov-roman:
before:
image
after:
image


also this scroll effect now applied correctly to both scroll to the bottom and to the top
image
e.g. this is before the change:
image

ilya-zlobintsev:
I'm not sure about the change of making the ScrollWindow global, it adds scrolling to views when there's no point of it. For example, there's no reason this should be scrollable:
image

Issue #718, opened 2025-08-19
Option to show clocks in mhz only instead of ghz


xnuken:
Would be nice to have a option to only show clocks in mhz and not have it auto jump to ghz.

ilya-zlobintsev:
Can you explain why you want this? The stats are specifically always shown as GHz to have a static width of the text (so it doesn't jump around when going below or above 1000 MHz).

xnuken:
Most software that is used for overclocking will show it in MHz instead of GHz or at least allowing the option to switch. It would just be a nice quality of life and option to have. Also the graph already shows MHz so would add a little consistency to it too.

ilya-zlobintsev:
The readings are MHz now.

PR #883, opened 2026-01-27
feat: refactor formatters and add gui tests

makarov-roman:
I felt kinda weird when I saw one widget depends on another, and then we import staff from app.rs
so this change is mostly moving reusable formatters into the formatting.rs file and adding tests for them.

But there are 2 changes in behaviour:
1. use Mb instead of Gb for VRAM.
the reason is that when we us GB - people don't understand why it's 7.9 or 11.9 instead of 8 or 12 and think it's a bug on our side. Of course we can add a tooltip explaining that "actually ...", but I think showing them in Mb solves the issue.
image
nvidia does it too

2. use mHz instead of GhZ for core/vram clocks
as mentioned in #718 it is a standard to show in mHz, also we're using mHz in OC widget, so it makes sense to use the same unit. After the visuals change in the last release it doesn't wiggle that much

### DIFF
--- a/lact-gui/src/app/formatting.rs
+++ b/lact-gui/src/app/formatting.rs
@@ -162,8 +162,8 @@ pub fn fmt_human_bytes(bytes: u64, unit: Option<ByteUnit>) -> String {
         (size, ByteUnit::ALL[i])
     };
 
-    let label = unit.label();
-    format!("{size:.1$} {}", label, (size.fract() != 0.0) as usize)
+    let size_str = format!("{:.1}", size).trim_end_matches(".0").to_string();
+    format!("{} {}", size_str, unit.label())
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


lact-gui/src/lib.rs
```diff
@@ -26,8 +28,16 @@ const GUI_VERSION: &str = env!("CARGO_PKG_VERSION");
...
-    LazyLock::new(|| i18n::loader(fluent_language_loader!(), &Localizations));
+pub(crate) static I18N: LazyLock<FluentLanguageLoader> = LazyLock::new(|| {
+    #[cfg(test)]
+    {
+        std::env::set_var("LANGUAGE", "en");
+        std::env::set_var("LC_ALL", "en_US.UTF-8");
+        std::env::set_var("LANG", "en_US.UTF-8");
+    }
+
+    i18n::loader(fluent_language_loader!(), &Localizations)
```
ilya-zlobintsev: Please use anything other than `set_var`. It is extremely unsafe in multi-threaded programs: https://doc.rust-lang.org/std/env/fn.set_var.html#safety, and Rust tests happen to run multi-threaded. The only reason it's marked a safe function in the standard lib is backwards compat (in newer editions of Rust it's marked unsafe).

Instead of this there should probably be a version of `fn loader` for tests, which hardcodes a specific language rather than loading the system one.

PR #884, opened 2026-01-27
fix: deduplicate vulkan device

makarov-roman:
fixes: https://github.com/flatpak/flatpak/issues/6418


makarov-roman:
tested with `make install-debug` and locally built flatpak
image

PR #892, opened 2026-02-01
feat: support multi-column layout for adjustment-rows lists

makarov-roman:
I think we should switch to album orientation later, but I don't like yet how other parts look.
image

### DIFF
--- a/lact-gui/src/app/pages/oc_page/clocks_frame.rs
+++ b/lact-gui/src/app/pages/oc_page/clocks_frame.rs
@@ -605,7 +605,6 @@ impl ClocksFrame {
     }
 
     pub fn get_commands(&self) -> Vec<SetClocksCommand> {
-        // If nvidia options are enabled, we always set locked clocks to None or Some
         self.all_groups()
             .flat_map(|group| group.get_commands())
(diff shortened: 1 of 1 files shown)


ilya-zlobintsev (changes requested):
This currently looks quite awkward on AMD with the separator lines, e.g. 7900xtx data:

Image

Perhaps the separators shouldn't be shown at the start/end if there are no items there.

It also changes the item order due to the items being in separate widgets, it used to be GPU clock -> VRAM clock -> GPU voltage, and now it's GPU clock -> GPU voltage -> VRAM clock

makarov-roman:
> This currently looks quite awkward on AMD with the separator lines, e.g. 7900xtx data:
> Image
> 
> Perhaps the separators shouldn't be shown at the start/end if there are no items there.
> 
> It also changes the item order due to the items being in separate widgets, it used to be GPU clock -> VRAM clock -> GPU voltage, and now it's GPU clock -> GPU voltage -> VRAM clock

it's good that you use kde :)
I literally see these separators for the first time.

nvm, see it now

makarov-roman:
image

ilya-zlobintsev (changes requested):

Issue #894, opened 2026-02-01
Improve power states list UI


ilya-zlobintsev:
- It does not make sense to show checkmarks next to the p-states on devices where they are never configurable (nv and intel)

<img width="719" height="214" alt="Image" src="https://github.com/user-attachments/assets/98c543f8-2eb6-4d69-ad9a-6e7653bc6e10" />

- Empty lists should not be shown (like in #893)

PR #896, opened 2026-02-02
feat: improve pstates UI

makarov-roman:
addresses #894
image
changes:
1) hide checkboxes instead of disabling them
2) do not render list when there are no values
3) move power_state_row to dedicated file (maybe it's my habbit, but I don't really like to have multiple components in one file)
4) remove global `sensitive` flag, instead all inactive pstates will be dimmed via default css class.
5) move active arrow label to the left to make it more obvious where it points ( if you stretch all UI on big screen it might be hard to tell). Also it prevents list width from jumping by couple of pixels.
6) add PageSectionExpander component and wrap pstates frame into that, it's copy-paste of PageSection mostly. I've tried to implement it first on PageSection, but it was looking very hacky.
7) switch to monospace to improve consistency

### DIFF
--- a/lact-gui/src/app/pages/oc_page/power_states/power_states_frame.rs
+++ b/lact-gui/src/app/pages/oc_page/power_states/power_states_frame.rs
@@ -72,6 +72,8 @@ impl relm4::SimpleComponent for PowerStatesFrame {
                     add_binding: (&model.states_configured, "active"),
                     #[watch]
                     set_visible: model.performance_level.is_some(),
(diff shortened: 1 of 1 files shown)


makarov-roman:
> Also, is it expected that the expander button is a bit out of bounds compared to the rest of the UI?

I wanted to align headers themselves
image

makarov-roman:
fixed 
image

ilya-zlobintsev (approved):
LGTM

PR #900, opened 2026-02-07
feat: calculate max clocks

makarov-roman:
I decided to move it to a dedicated PR to make it easier to review.


lact-daemon/src/server/gpu_controller/amd.rs
```diff
@@ -498,10 +498,34 @@ impl AmdGpuController {
                 .map(u64::from);
         }
 
+        let clocks_table = self.handle.get_clocks_table().ok();
```
makarov-roman: is it fine to call for `handle.get_clocks_table` here? I didn't find usage for it up in the `get_stats`.

makarov-roman: overall the implementation based on clocks_table seems to be much cleaner than p-states based.

PR #903, opened 2026-02-07
feat: improve gpu stats UI

makarov-roman:
I wasn't happy how it looks like when LevelBar and text are mixed together. 
Also I think it's better to show some metrics via LevelBar.

image


ilya-zlobintsev (changes requested):
Perhaps it makes sense to force minimum 2 columns of stats, or at least adjust the reflow threshold? It looks like this at the default window size with breeze, which is very inefficient:

Image

Also, I'm not sure about the fact that now the stats are split into two sections. Maybe the non-"bar" section should be first, I think it doesn't make sense that the GPU name is in the middle.

lact-gui/src/app/pages.rs
```diff
@@ -5,11 +5,12 @@ pub mod oc_page;
...
 
-use lact_schema::{DeviceInfo, DeviceStats};
+use lact_schema::{DeviceInfo, DeviceStats, PowerStates};
 use std::sync::Arc;
 
 #[derive(Debug, Clone)]
 pub enum PageUpdate {
     Info(Arc<DeviceInfo>),
     Stats(Arc<DeviceStats>),
+    PowerStates(Arc<PowerStates>),
```
ilya-zlobintsev: I think instead of adding this to the generic `PageUpdate`, it's better to make a separate message type for the stats section, as other places using this type don't really care about the power states.

ilya-zlobintsev (changes requested):
While this hasn't been changed in this PR, I think the formatting for VRAM usage should be changed to have at least one (maybe two) decimal places. Currently, it is not as useful as it would be with more accurate data IMO:

Image

makarov-roman:
> While this hasn't been changed in this PR, I think the formatting for VRAM usage should be changed to have at least one (maybe two) decimal places. Currently, it is not as useful as it would be with more accurate data IMO:

@ilya-zlobintsev I just added new test case which shows that 90mb is displayed as 0.1 Gb. Which means the problem is with something else and VRAM usage value is literally 0 in the state. Is it on your machine or some snapshot?

ilya-zlobintsev:
> > While this hasn't been changed in this PR, I think the formatting for VRAM usage should be changed to have at least one (maybe two) decimal places. Currently, it is not as useful as it would be with more accurate data IMO:
> 
> @ilya-zlobintsev I just added new test case which shows that 90mb is displayed as 0.1 Gb. Which means the problem is with something else and VRAM usage value is literally 0 in the state. Is it on your machine or some snapshot?

This is on a headless system with a very low, but non-zero VRAM usage:
```
VRAM Usage: 17/6088 MiB
```
image

FreshChriss:
I like the idea of dividing the bars from the text only values. The thing i am missing is the fan steed in % (as a bar or just text value, i think bar would make more sense tho). It was added with the last public release (by you, thanks again for that).

I like to keep track of both values because on my nvidia card the % value is more useful (but the rpm is still interesting) and on my amd gpu its the other way around.

Greetings
Chriss

makarov-roman:
> VRAM Usage: 17/6088 MiB

I would argue that we should round it to zero here :) 17 MB is nothing with modern HW.
- 0.1 GB would be less correct.
- 17 MiB would mean that we have to use a dynamic unit, which is easy to do (the formatter supports that), but we'll have a lot of layout changes near breaking points, e.g. 990 MiB / 1 GB

makarov-roman:
> I like the idea of dividing the bars from the text only values. The thing i am missing is the fan steed in % (as a bar or just text value, i think bar would make more sense tho). It was added with the last public release (by you, thanks again for that).

Hey, my idea is that the LevelBar represents %. It's a bit strange with this UI to show them both. The screenshot does show 0 RPM, but the LevelBar represents 0%. Hope it makes sense.

ilya-zlobintsev (approved):

FreshChriss:
> > I like the idea of dividing the bars from the text only values. The thing i am missing is the fan steed in % (as a bar or just text value, i think bar would make more sense tho). It was added with the last public release (by you, thanks again for that).
> 
> Hey, my idea is that the LevelBar represents %. It's a bit strange with this UI to show them both. The screenshot does show 0 RPM, but the LevelBar represents 0%. Hope it makes sense.
> 

Sorry for the really late reply. I understand what you are doing, but in my case i am talking about the difference between 15-20% (thats the fan operating range for my 9070 xt), and i dont think i am able to make out the exact values in between by the bar alone (like, is my gpu at 17% or 18% fan, because thats actually how sensitive (or insensitive) this model is, and thats useful information for me). Same with my nvidia gpu that is usually between 30-36% and each 1% makes it significantly louder.

I would suggest adding the fan% as a value next to the bar and putting the rpm in the text only section? (The other way around is also fine for me personally, i would just like to keep both values visible as a number, preferably still together as it is in the current release but i get why the bar is also a neat addition for that value)

Greetings
Chriss

makarov-roman:
Hi @FreshChriss,
that seems like a valid reason to me, please open an issue

FreshChriss:
> Hi @FreshChriss,
> that seems like a valid reason to me, please open an issue

Hi @makarov-roman,
Thanks for the reply.
As requested i have opened an issue for this. Its my first time creating an issue on github so im not sure it i have done it correctly, please excuse potential mistakes because of that.

The issue should be here:
https://github.com/ilya-zlobintsev/LACT/issues/950

PR #911, opened 2026-02-15
feat: add level bar animation

makarov-roman:
https://github.com/user-attachments/assets/728ea232-a009-47c2-a6a7-5d7304d6b424


makarov-roman:
would look nicer with #903 
if we migrate on adw, then this animation can be replaced

makarov-roman:
merged master to see how it looks with new gpu-stats and found a bug. Decided to switch to linearAnimation which is as simple as animation can be.

https://github.com/user-attachments/assets/8a29bb63-efa9-477f-9198-7eb2d51ab805

makarov-roman:
such a tiny change with adw

ilya-zlobintsev (approved):
Looks good

PR #926, opened 2026-02-27
set /etc/lact to 0600 on creation

sofoxe1:


sofoxe1:
how about setting it to root:admin group or admin user:admin group? imo apps which wouldn't be able to interact with lact daemon have no reason to read the config

ilya-zlobintsev:
> imo apps which wouldn't be able to interact with lact daemon have no reason to read the config

System-wide configs are generally world-readable, excluding security-sensitive ones, which the lact config isn't. You can just check `/etc` on any system, most things will be 644 or similar. I don't see a reason why lact should be different.

PR #927, opened 2026-02-28
pkg: bump minimum deb distro versions

ilya-zlobintsev:


Ph0rk0z:
Any way to keep 22.04? Build from source? Not so simple for me to just upgrade as I have a server.

ilya-zlobintsev:
> Any way to keep 22.04? Build from source? Not so simple for me to just upgrade as I have a server.

Use flatpak or docker (the latter only if you don't need GUI).

Ph0rk0z:
Heh.. I use neither of those. Docker eats resources and so does flatpack. GUI I actually do use to set profiles and check on GPUs.

PR #931, opened 2026-02-28
feat: enable adwaita by default, add theme selector and detection

makarov-roman:


ilya-zlobintsev:
~~TODO: figure out broken look in cosmic with dark theme (and potentially with other `colors.css` customizations)?~~
EDIT: fixed
image

rcv11x:
Sorry for commenting on this PR, but does this affect KDE users? I mean, will we be seeing the Adwaita theme now?

makarov-roman:
@rcv11x
np, basically yes, but we still want to keep styling coherent with the rest of the system where possible.

KDE users will see adwaita with breeze theme. Currently it looks like this:
image

plain gtk is lacking, and only KDE had a decent default theme. Which means that LACT was consistent with the rest of the system only on KDE.

Switching to adwaita allows us to use their UI components, thus we can focus more on core functionality instead of implementing basic UI elements.

ilya-zlobintsev:
> Sorry for commenting on this PR, but does this affect KDE users? I mean, will we be seeing the Adwaita theme now?

Basically what was mentioned already - LACT will now look slightly different, but it will still attempt to integrate with the system when possible, including on KDE. You will still have breeze colours for example.

PR #944, opened 2026-03-20
feat: restrict content max width

makarov-roman:
image
image
image
image
image
curve is big enough, I don't making it bigger improves usability

I specifically decided not to touch header and applyRevealer because it will be moved to the sidebar short term

### DIFF
--- a/lact-gui/src/app.rs
+++ b/lact-gui/src/app.rs
@@ -133,13 +133,14 @@ impl AsyncComponent for AppModel {
                 gtk::Box {
                     set_orientation: gtk::Orientation::Vertical,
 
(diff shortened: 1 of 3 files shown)


ilya-zlobintsev (changes requested):
I initially didn't like it from the screenshots, but actually using it - this feels nice

Scrolling seems broken however when the window is small:

Image

makarov-roman:
@ilya-zlobintsev :man_facepalming: thanks for catching that. fixed

ilya-zlobintsev (approved):
Using the app maximized made me realize that the apply/revert popover maybe isn't positioned the best, this can also be changed with the sidebar

PR #957, opened 2026-04-14
feat: add Nvidia VF curve editor

### LINKED ISSUE #936: Feature: Per-point V/F curve reading and editing for NVIDIA GPUs via undocumented NvAPI
# Feature: Per-point V/F curve reading and editing for NVIDIA GPUs on Linux

## Summary

I've discovered undocumented NvAPI functions accessible via `libnvidia-api.so` on Linux that enable **full 128-point V/F (voltage-frequency) curve reading and per-point frequency offset control** for NVIDIA GPUs — the same mechanism MSI Afterburner uses on Windows.

This was previously considered impossible on Linux (see #847, #453). The functions are confirmed working on an RTX 5090 with driver 590.48.01.

### 1. Read full V/F curve (128 points)

Function `ClockClientClkVfPointsGetStatus` (`0x21537AD4`) returns all 128 voltage-frequency pairs:

```
Point  Frequency  Voltage
  0     225 MHz    450 mV    (idle)
 51     247 MHz    770 mV
 70    1875 MHz    890 mV
 80    2640 MHz    950 mV
 90    2850 MHz   1015 mV
100    2977 MHz   1075 mV
110    3090 MHz   1140 mV
120    3187 MHz   1200 mV
126    3232 MHz   1240 mV    (max boost)
127     405 MHz    530 mV    (low-power)
```

### 2. Set per-point frequency offsets

Function `ClockClientClkVfPointsSetControl` (`0x0733E009`) sets a frequency offset (in kHz) for individual V/F points. This enables true curve reshaping — different offsets at different voltage levels — unlike the global `nvmlDeviceSetGpcClkVfOffset` which shifts the entire curve uniformly.

**Tested and confirmed:** setting +20 MHz offset on individual points succeeds, affects actual GPU boost behavior (voltage changes observed), and can be reset to 0.

### 3. Read current offsets and allowed ranges

- `ClockClientClkVfPointsGetControl` (`0x23F1B133`) — read current per-point offsets
- `ClockClientClkDomainsGetInfo` (`0x64B43A6A`) — read allowed offset range (±1000 MHz for GPU core, -1000/+3000 MHz for memory on RTX 5090)

## Discovered NvAPI Functions

All accessed via `libnvidia-api.so` → `nvapi_QueryInterface(function_id)`.
Version field format: `(version_number << 16) | struct_size`.

### Read-only (safe)

| Function ID | Name | Struct Size | Description |
|---|---|---|---|
| `0x21537AD4` | ClkVfPointsGetStatus | 0x1C28 (7208) | Full 128-point V/F curve |
| `0x507B4B59` | ClkVfPointsGetInfo | 0x182C (6188) | Active point mask |
| `0x23F1B133` | ClkVfPointsGetControl | 0x2420 (9248) | Current frequency offsets per point |
| `0x64B43A6A` | ClkDomainsGetInfo | 0x0928 (2344) | Clock domain ranges (allowed offset min/max) |
| `0xE440B867` | PerfClientLimitsGetStatus | 0x030C (780) | V/F lock status |
| `0x465F9BCF` | ClientVoltRailsGetStatus | 0x004C (76) | Current GPU voltage (already in LACT) |
| `0x9DF23CA1` | ClientVoltRailsGetControl | 0x0028 (40) | Voltage boost % |
| `0x6FF81213` | GetPstates20 | 0x1C94 (7316) | Performance states |

### Write (tested on RTX 5090)

| Function ID | Name | Description |
|---|---|---|
| `0x0733E009` | ClkVfPointsSetControl | Set frequency offset per V/F point |

### ClkVfPointsGetStatus (`0x21537AD4`)

```
Offset  Size  Field
0x00    4     version = 0x00011C28
0x04    16    mask (128 bits — set all 0xFF to request all points)
0x14    4     numClocks = 15
0x18    48    padding/header
0x48    3584  128 × 28-byte entries
```

Each entry (28-byte stride):
```
+0x00   4   frequency_kHz (uint32)
+0x04   4   voltage_µV (uint32)
+0x08   20  reserved
```

### ClkVfPointsSetControl (`0x0733E009`)

```
Size: 0x2420, Version: 1
0x00    4     version = 0x00012420
0x04    16    mask — set ONLY the bit for the point you're modifying
0x20    9216  128 × 72-byte entries (stride 0x48)
```

Each entry:
```
+0x00   4   frequency_offset_kHz (int32, signed)
```

**Critical detail:** The mask must have only ONE bit set per call. Setting all 128 bits (0xFF×16) causes error -1. Call the function once per point with a single-bit mask.

## Important Caveat: NVML Conflict

**NvAPI SetControl and NVML `nvmlDeviceSetGpcClkVfOffset` operate on the same underlying hardware state and conflict with each other.** When both are used:
- NvAPI writes can clear NVML-applied offsets
- The interaction is unpredictable

LACT would need to choose one path:
- **Option A:** Use NvAPI for per-point V/F control, stop using NVML for clock offsets
- **Option B:** Use NvAPI read-only (display the curve in GUI), keep NVML for writes
- **Option C:** Let the user choose between "simple mode" (NVML global offset) and "advanced mode" (NvAPI per-point)

## PoC Code

Working Python proof-of-concept for reading and writing: https://gist.github.com/Loong0x00/0163c73d9379d6e683d83c2718791048

Minimal read example:
```python
import ctypes, struct

nvapi = ctypes.CDLL("libnvidia-api.so")
QI = nvapi.nvapi_QueryInterface
QI.restype = ctypes.c_void_p
QI.argtypes = [ctypes.c_uint32]

def nvfunc(fid, nargs=2):
    p = QI(fid)
    if not p: return None
    return ctypes.CFUNCTYPE(ctypes.c_int32, *[ctypes.c_void_p]*nargs)(p)

# Initialize
nvfunc(0x0150E828, 0)()  # NvAPI_Initialize
gpus = (ctypes.c_void_p * 64)()
ngpu = ctypes.c_int32()
nvfunc(0xE5AC921F)(ctypes.byref(gpus), ctypes.byref(ngpu))
gpu = gpus[0]

# Read V/F curve
buf = ctypes.create_string_buffer(0x1C28)
struct.pack_into("<I", buf, 0, (1 << 16) | 0x1C28)
for i in range(4, 20): buf[i] = 0xFF  # mask: all 128 points
struct.pack_into("<I", buf, 0x14, 15)   # numClocks
nvfunc(0x21537AD4)(gpu, buf)
d = bytes(buf)

for i in range(128):
    freq = struct.unpack_from("<I", d, 0x48 + i * 0x1C)[0]      # kHz
    volt = struct.unpack_from("<I", d, 0x48 + i * 0x1C + 4)[0]  # µV
    if freq > 0:
        print(f"Point {i}: {freq/1000:.0f} MHz @ {volt/1000:.0f} mV")
```

Minimal per-point write:
```python
# Set +50 MHz offset on point 80
set_func = nvfunc(0x0733E009)
buf = ctypes.create_string_buffer(0x2420)
struct.pack_into("<I", buf, 0, (1 << 16) | 0x2420)
buf[4 + 80 // 8] = 1 << (80 % 8)  # single-bit mask for point 80
struct.pack_into("<i", buf, 0x20 + 80 * 0x48, 50000)  # +50 MHz in kHz
ret = set_func(gpu, buf)  # returns 0 on success
```

## How this was discovered

Reverse engineering of ASUS GPU Tweak III `Vender.dll` using Ghidra, cross-referenced with:
- [nvapi-rs](https://github.com/arcnmx/nvapi-rs) (function ID definitions)
- [nvapioc](https://github.com/Demion/nvapioc) (confirmed V/F curve IDs)
- [vertminer-nvidia](https://github.com/vertcoin-project/vertminer-nvidia) (ID defines)

NvAPI error codes on Linux use negative integers (not `0x80000000+` range): `-9` = `INCOMPATIBLE_STRUCT_VERSION`, `-1` = generic error.

## Hardware Tested

- GPU: ASUS ROG Astral RTX 5090 (Blackwell, GB202)
- Driver: 590.48.01
- OS: Arch Linux, kernel 6.19.6

## Suggested Implementation

For LACT, the lowest-risk first step would be **Option B** — read-only V/F curve display:
1. Add the `GetStatus` call to `nvapi.rs` (extends existing NvAPI infrastructure)
2. Display the 128-point curve in the OC page GUI
3. Show actual voltage vs frequency mapping — users currently have no visibility into this

Per-point write support could follow once the read path is validated across more GPU models.

## Related Issues

- #847 — "NVIDIA curve frequency" (closed as not possible — this issue shows it IS possible)
- #453 — "Curve optimizer" (closed, same reason)
- #486 — Current NVML-based pseudo-undervolt workaround
- #625 — Existing NvAPI voltage reading (same infrastructure)

ilya-zlobintsev:
Closes #936 

This PR adds a VF curve editor, based on some undocumented NvAPI queries. There are no guarantees about its safety, stability across driver versions, etc. But it might still be useful for some people who might want to tweak this at their own risk.
With this editor, you can do undervolts in a very similar way to MSI Afterburner on Windows, by bumping a desired voltage point to the desired clockspeed, and flattening the curve to the right:

Special thanks to @Ristovski for helping with the reverse engineering efforts

image

### DIFF
--- a/lact-daemon/src/server/gpu_controller/nvidia.rs
+++ b/lact-daemon/src/server/gpu_controller/nvidia.rs
@@ -5,7 +5,11 @@ use super::{CommonControllerInfo, FanControlHandle, GpuController};
 use crate::{
     bindings::nvidia::NvPhysicalGpuHandle,
     server::{
-        gpu_controller::{NvApi, common::fan_control::FanCurveExt, common::resolve_process_name},
+        gpu_controller::{
+            NvApi,
+            common::{fan_control::FanCurveExt, resolve_process_name},
+            nvidia::nvapi::{CLOCK_CLIENT_CLK_VF_POINT_TYPE_PROG, ClockClientClkVfPointInfoV1},
+        },
         opencl::get_opencl_info,
         vulkan::get_vulkan_info,
     },
@@ -21,10 +25,10 @@ use indexmap::IndexMap;
 use lact_schema::{
     CacheInfo, ClocksInfo, ClocksTable, ClockspeedStats, DeviceFlag, DeviceInfo, DeviceStats,
     DeviceType, DrmInfo, DrmMemoryInfo, FanControlMode, FanStats, IntelDrmInfo, LinkInfo,
-    NvidiaClockOffset, NvidiaClocksTable, PmfwInfo, PowerState, PowerStates, PowerStats,
-    ProcessInfo, ProcessList, ProcessType, ProcessUtilizationType, TemperatureEntry, VoltageStats,
-    VramStats,
-    config::{FanControlSettings, FanCurve, GpuConfig},
+    NvidiaClockOffset, NvidiaClocksTable, NvidiaVfPoint, PmfwInfo, PowerState, PowerStates,
+    PowerStats, ProcessInfo, ProcessList, ProcessType, ProcessUtilizationType, TemperatureEntry,
+    VoltageStats, VramStats,
+    config::{CurvePoint, FanControlSettings, FanCurve, GpuConfig},
 };
 use nvml_wrapper::{
     Device, Nvml,
@@ -53,13 +57,12 @@ const SUPPORTED_UTIL_TYPES: &[ProcessUtilizationType] = &[
 
 pub struct NvidiaGpuController {
     nvml: &'static Nvml,
-    nvapi: Option<&'static NvApi>,
     common: CommonControllerInfo,
     fan_control_handle: RefCell<Option<FanControlHandle>>,
     initial_target_temp: Option<u32>,
 
+    nvapi: Option<(&'static NvApi, NvPhysicalGpuHandle)>,
     driver_handle: Option<DriverHandle>,
-    nvapi_handle: Option<NvPhysicalGpuHandle>,
     nvapi_thermals_mask: Option<i32>,
 
     last_util_timestamp: Cell<Option<u64>>,
@@ -67,6 +70,8 @@ pub struct NvidiaGpuController {
     last_applied_offsets: RefCell<HashMap<Clock, HashMap<PerformanceState, i32>>>,
     last_applied_gpu_locked_clocks: RefCell<Option<(u32, u32)>>,
     last_applied_vram_locked_clocks: RefCell<Option<(u32, u32)>>,
+    // Check if reset is needed to avoid unnecessarily going to nvapi
+    vf_curve_written: Cell<bool>,
 }
 
 impl NvidiaGpuController {
@@ -130,10 +135,9 @@ impl NvidiaGpuController {
 
         Ok(Self {
             nvml,
-            nvapi,
+            nvapi: nvapi.zip(nvapi_handle),
             common,
             driver_handle,
-            nvapi_handle,
             nvapi_thermals_mask,
             initial_target_temp: target_temp,
             last_util_timestamp: Cell::new(None),
@@ -141,6 +145,7 @@ impl NvidiaGpuController {
             last_applied_offsets: RefCell::new(HashMap::new()),
             last_applied_gpu_locked_clocks: RefCell::new(None),
             last_applied_vram_locked_clocks: RefCell::new(None),
+            vf_curve_written: Cell::new(false),
         })
     }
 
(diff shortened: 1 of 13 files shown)


Panchovix:
This is amazing, incredible work!

Ph0rk0z:
Unlike with locked clocks, doing it this way makes the frequency jump around a bit. I see from 1660-1710. The value of the offset to maintain frequency goes up and down even if you are in the same pstate.

jpietek:
That's a major milestone to make the Nvidia actually useable on Linux, thank you for that!

Funy story is that for the last few days I tried to use Codex to brute force reverse engineer Nvidia libs for the same purpose. Had some success with nvml library, for example got the voltage getters.

But had a big struggle to find out those vf curve control setters in nvapi library, and eventually stumbled upon this merge pull request to have a ready made solution.

My idea for the tool was to have a single cli line importer for MSI AfterBurner 4.6 profiles. You can take a look at it here. If at some point the import feature would be beneficial for the LACT user, feel free to have some inspiration.

https://github.com/jpietek/PenguinBurner

Personally I have Founders Edition 5080 which ran like total crap in terms of efficiency, temperatures and noise on Linux. Did years of the ugly workarounds with power limits, constant frequency and static voltage offset, but the AfterBurner heavily tuned curves on Windows where so much better. Well, not anymore! :)

taosecurity:
This is wonderful work. I'd love to see the MSI Afterburner import feature in LACT too.

jpietek:
I've added LACT export today, so now you can bridge those tools together or import a curve from my algorithmic auto UV.

FlightSuitRiot:
Is there a way we can get the keystroke functionality of the Afterburner V/F curve too? Ex: hold shift and drag cursor to select batch of v/f curve points, hold Enter to move them all up or down at once?

makarov-roman:
@FlightSuitRiot please create an issue with feature request and explain what do you want to achieve. Do I get it right that you want to select multiple volt points and apply core offset on them?

Also please explain what you're trying to achieve. Coping UX from Afterburner sometimes makes sense, because people are somewhat familiar with it, sometimes doesn't, because it generally bad.
So we might find a better way to achieve what you want.

PR #964, opened 2026-04-18
feat: sync selected page title into content's Toolbox header

makarov-roman:
image
should we rename information to hardware info to make it less generic?
image

### DIFF
--- a/lact-gui/i18n/en/lact_gui.ftl
+++ b/lact-gui/i18n/en/lact_gui.ftl
@@ -1,5 +1,5 @@
 info-page = Information 
-oc-page = OC 
+oc-page = Overclock
 thermals-page = Thermals 
 software-page = Software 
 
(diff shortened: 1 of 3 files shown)


lact-gui/i18n/en/lact_gui.ftl
```diff
@@ -1,5 +1,5 @@
 info-page = Information 
-oc-page = OC 
+oc-page = Overclock
```
ilya-zlobintsev: Maybe it would be more consistent to call it `Overclocking`? Other pages are nouns, and this reads like a verb (even if it can be both)

ilya-zlobintsev:
> information to hardware info to make it less generic? 

I'm afraid it might either not fit in the sidebar, or make the sidebar too wide, or just look bad compared to other short page names

makarov-roman:
> > information to hardware info to make it less generic?
> 
> I'm afraid it might either not fit in the sidebar, or make the sidebar too wide, or just look bad compared to other short page names

it's fine. It's 1 character more than Overclocking. And I expect that GPU picker will always dictate the min width

image

ilya-zlobintsev (approved):
It looks fine actually, the text is not as long as I thought it would be.

PR #970, opened 2026-04-19
chore: update PageSection api to handle better hide_visible_container prop

makarov-roman:
image
image

### DIFF
--- a/lact-gui/src/app/page_section.css
+++ b/lact-gui/src/app/page_section.css
@@ -3,3 +3,11 @@
 .page-section-children-box {
     margin: 10px;
 }
+
+.page-section-no-container .page-section-children-box {
+    margin: 0;
+}
+
+.page-section-no-container .boxed-list row {
+    padding: 10px;
+}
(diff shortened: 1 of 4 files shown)


ilya-zlobintsev:
I'm not a fan of this. The cache sizes are probably irrelevant information to 99% of people (it was added because I thought it would be cool to report, but practically I don't think anyone cares about it). Yet with this change, it takes up a significant part of the UI, almost as much as the rest of the text:
image

I think it should be hidden behind some popup/dropdown, the current display can definitely be improved, but this is too verbose IMO.

makarov-roman:
@ilya-zlobintsev it makes sense. I agree. the cache info is not that useful to dedicate so much space to it. 

during this PR I've made a small refactoring to PageSection, which helps to use it without default card container, I'll rollback the cache block so we can merge the small improvement alone.

https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.9/widget-gallery.html#expander-row can be a solution though

PR #971, opened 2026-04-19
fix: fix incorrect scrollbar position when clamped

makarov-roman:
before:
image
after:
image

### DIFF
--- a/lact-gui/src/app.rs
+++ b/lact-gui/src/app.rs
@@ -395,6 +395,7 @@ impl AsyncComponent for AppModel {
 
                                     adw::Clamp {
                                         set_maximum_size: CONTENT_MAXIMUM_WIDTH,
+                                        set_tightening_threshold: CONTENT_MAXIMUM_WIDTH,
 
                                         #[name = "root_stack"]
                                         gtk::Stack {
--- a/lact-gui/src/app.rs
+++ b/lact-gui/src/app.rs
@@ -139,7 +139,7 @@ impl AsyncComponent for AppModel {
         #[root]
         adw::ApplicationWindow::builder()
             .default_height(750)
-            .default_width(1000)
+            .default_width(1300)
             .icon_name(APP_ID)
             .title("LACT")
             .build() {
--- a/lact-gui/src/app.rs
+++ b/lact-gui/src/app.rs
@@ -139,7 +139,7 @@ impl AsyncComponent for AppModel {
         #[root]
         adw::ApplicationWindow::builder()
             .default_height(750)
-            .default_width(1300)
+            .default_width(1100)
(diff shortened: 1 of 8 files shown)


makarov-roman:
@ilya-zlobintsev I decided to add some changes with spacing here, so it can be tested in one place. 
After the sidebar was introduced, I don't really like how the default layout look like. E.g. in adjustment row. 

I decreased the default margins (also removed the bottom margin which was used for the apply revealer) to fit more content when the app is not fullscreen. And increased default width so the oc adjustment rows are forced to 2 rows on most of the configs with 100% font scaling

ilya-zlobintsev (commented):
No problems with the spacing and scrollbar changes, but do we really need to increase the default window width? I found that it doesn't universally improve things, e.g. here there's not really any extra information displayed compared to the old size:

Image

Image

I would generally prefer to keep the default window size compact, and let people increase them as they want. Also, it's possible to add window size persistency across launches, but I'm not sure if that's worth it

makarov-roman:
@ilya-zlobintsev I think it's worth do add persistency. But as a separate PR.

regarding the window size - I want the default to have the best layout we have, so I was trying to force 2 columns on adjustment rows because they look much better this way. We somehow lost the sweet spot with 2 columns for gpu stats and 2 columns with adjustment rows.

If I would tune the UI just for myself, I'd remove scale-bar in the adjustment rows, just an input with buttons should be more enough to set the values conveniently. Then the compact width should look good too.

makarov-roman:
@ilya-zlobintsev I checked the old layouts (pre sidebar, pre adw) and we never had 2 columns adj-row and and 2 columns stats by default, it's either one or another.

I set the width to 1100 (900 as we had before + 200 sidebar). I think it's good, but I really have no preference atm. So feel free to merge as is, or rollback the width and merge

regarding adjustment-row, I still want to remove the scale from where, I think +- and input is enough :) this will also give us 2 columns in the default layout. let me know what you think, if you agree, I'll make another PR

ilya-zlobintsev:
> regarding adjustment-row, I still want to remove the scale from where, I think +- and input is enough :) this will also give us 2 columns in the default layout. let me know what you think, if you agree, I'll make another PR

I disagree actually. With scales you can immediately see how far the current value is from the possible min/max (for example in locked clocks on nvidia). Without it, you would have to try actually setting the values to know the possible range, or have separate text for describing it. In general, I think a slider is a nice control here, though this is probably subjective.

We could look for ways to make the adjustment rows more compact though, like maybe shortening the text

makarov-roman:
@ilya-zlobintsev might be not something we have to resolve, since we're splitting the page anyways

PR #972, opened 2026-04-19
feat/rework preferences popover

makarov-roman:
changes:
1. new preferences dialog (plain adw). disable AMD OC moved there. I suggest to add language switcher (when somebody doesn't want it to follow system language), daemon log level e.t.c. but later, imo it's fine to keep it simple for now
image
2. update interval moved to sidebar. With the idea to have stats/metrics on multiple pages it makes more sense
image
3. add plain adw about dialog with some useful links. Just for consistency with ecosystem
image

the popover now looks like this:
image


ilya-zlobintsev (commented):
Some thoughts:

- I like having a settings dialog in general
- I don't think the update interval configuration should be in the sidebar. It's something most people don't need to touch, and I think it's a good candidate to be in the (currently kinda empty) settings dialog. It especially doesn't make sense for it to always be there on e.g. the info page
- `about_dialog` can probably be inline in `app.rs`, it's a single widget
- Clicking on settings/about from the menu should close the menu automatically

makarov-roman:
> don't think the update interval configuration should be in the sidebar. It's something most people don't need to touch, and I think it's a good candidate to be in the (currently kinda empty) settings dialog. It especially doesn't make sense for it to always be there on e.g. the info page

when you OC and monitor I'd use different interval (e.g. when I game, I want it to use less resources, when I OC, I want to see the most relevant info). I though that it's nice, I'd use it from time to time, it's visible.

> about_dialog can probably be inline in app.rs, it's a single widget

my motivation is - I really don't like to make changes in app.rs right now, especially when working on multiple PRs which afffect it. I think it's perfectly fine to have component which is used once, if it allows to encapsulate all the logic related to the component. So for me it's not just a component, but a module.

Ideally I'd like to have app.rs as AppModel + mounts for the components

> Clicking on settings/about from the menu should close the menu automatically

fixed, I see that using menu model instead of set_popover will handle it automatically, I checked relm4 menu macro and it looks ugly (the logic has to be split all over the place), better manually closing it

ilya-zlobintsev:
> when you OC and monitor I'd use different interval (e.g. when I game, I want it to use less resources)

Even the minimum interval of 250ms is quite high, the resource usage should be negligible. If it's not for some reason, that's an issue that needs to be fixed

> Ideally I'd like to have app.rs as AppModel + mounts for the components

I agree that `app.rs` is quite huge, but my concern isn't really about reusability, but rather that there's a decent amount of boilerplate for defining and instantiating a component, when it's actually just 8 lines of actual widget code

> fixed, I see that using menu model instead of set_popover will handle it automatically, I checked relm4 menu macro and it looks ugly (the logic has to be split all over the place), better manually closing it

The menu originally used that, it's quite verbose to register and handle gtk "actions", but the main thing is that the macro wants static menu item names which is incompatible with localization

makarov-roman:
moved the interval to preferences

> I agree that app.rs is quite huge, but my concern isn't really about reusability, but rather that there's a decent amount of boilerplate for defining and instantiating a component, when it's actually just 8 lines of actual widget code

Until it isn't :) I can see that we might extend the dialog, and then it would be tempting to add more code to the app.rs to keep diff smaller. Imports and relm staff is a boilerplate indeed, but it's not the one which makes maintenance harder, IDE handles it semi-automatically

ilya-zlobintsev (approved):

PR #978, opened 2026-04-20
Use `jiff` instead of `chrono`

git-f0x:
Just ports stuff from chrono to jiff. Everything seems to work fine.

Is the date needed in the graphs bench? I removed it since it simplifies stuff.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -40,15 +40,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


ilya-zlobintsev (approved):
> Is the date needed in the graphs bench? I removed it since it simplifies stuff.

Doesn't really matter, as long as the timestamp is still increasing (which it is)

PR #983, opened 2026-04-23
feat: groom charts dialog

makarov-roman:
the idea is to get "easy gains" from adwaita, and fix some layout problems.

Notable change is that I removed root.fill for charts to make them transparent, thus we control the background size via it's containers. I saw that you're getting the colors from css variables, but then you still have to do more work to adopt it in each particular case.

legend looks a bit odd. I'm thinking about replacing it with some overlay component instead. (out of scope for now, I have no vision atm how to make it good). 

the diff is unnecessary big due to the changes to plot's container, but this is the price for a readable structure.

### DIFF
--- a/lact-gui/src/app/graphs_window/mod.rs
+++ b/lact-gui/src/app/graphs_window/mod.rs
@@ -87,7 +87,7 @@ impl relm4::Component for GraphsWindow {
                     set_orientation: gtk::Orientation::Horizontal,
                     set_spacing: 10,
                     set_margin_horizontal: 10,
-                    set_margin_vertical: 5,
+                    set_margin_top: 5,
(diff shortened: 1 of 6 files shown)


ilya-zlobintsev:
I don't quite like how this looks. The contrast was way more clear before:
image
And now:
image

Additionally, filling in the background colour has an effect of masking the fact that the `plotters` backend miscalculates the box size for the legend. Right now this ends up with text clipping out of it.

I also don't really like the overlayed editing buttons, since they overlap with text.

makarov-roman:
@ilya-zlobintsev removed the colours changes for now. ready for review

ilya-zlobintsev (approved):

PR #986, opened 2026-04-24
feat: groom profiles

makarov-roman:


ilya-zlobintsev (approved):
Minor nitpick: the `#[root]` attribute in `view!` is not needed as long as it is the only widget at the top level (also applies to the adwaita dialogs you added in other PRs)

ilya-zlobintsev:
> the #[root] attribute in view! is not needed 

Cleaned up in https://github.com/ilya-zlobintsev/LACT/commit/cd7899d125bfc44be76c9eaacb1d8fb21859388b

PR #989, opened 2026-04-26
feat: add 12VHPWR/12V-2x6 power connector monitoring via I2C

NeedForSpeed73:
## Summary
Adds real-time monitoring of the 12VHPWR/12V-2x6 power connector current 
for supported NVIDIA GPUs (currently confirmed on ASUS RTX 5090 Astral/Matrix).

## How it works
The connector exposes per-pin current and voltage data via SMBus/I2C 
(default address 0x2b, register base 0x80). This is configured per-GPU 
in the daemon config.

## Configuration
Add to `/etc/lact/config.yaml` under the GPU entry:
```yaml
gpus:
  <gpu-id>:
    power_connector:
      i2c_bus: N  # find with: sudo i2cdetect -l
```

## Features
- Per-pin current display in OC page (6 pins, 2-column layout)
- Total current and total power summary
- Historical graphs with appropriate 10A Y-axis limit
- Configurable alarm: threshold, desktop notification, custom command
- Full i18n support

## Hardware compatibility
Tested on ASUS RTX 5090 Astral OC. Should work on any GPU where the 
12VHPWR controller is accessible at I2C address 0x2b.
Known compatible: ASUS Astral, ASUS Matrix series.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -233,6 +233,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5d20789868f4b01b2f2caec9f5c4e0213b41e3e5702a50157d699ae31ced2fcb"
 
 [[package]]
+name = "byteorder"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
+
+[[package]]
 name = "bytes"
 version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


ilya-zlobintsev:
General thoughts: I'm somewhat reluctant to accept this functionality, because it is very niche and basically untestable without having one of the very few specific ASUS cards that expose this sensor. I'm not against it in principle, but it should be something that's well separated from the rest of the code and doesn't get in the way.

In terms of the implementation itself, I think the current approach where you have to specify the i2c bus manually makes this feature much less useful than it could be, since most of the people (even among the few that have a compatible card) will not know about the existence of this without explicitly digging through docs. Additionally, I'm not sure if i2c bus numbers are stable enough to be safely put into configuration. 
It would be much better if the bus was detected automatically on cards that are known to support it (this would need a whitelist of device ids, but that's likely what asus' first party software has as well).

Additionally, this PR appears to be vibecoded. I can give code feedback if there is a human reading it and understanding, but I don't want it to just end up being fed into an LLM with no thought given to it.

tolga9009:
> Additionally, this PR appears to be vibecoded.

Seems to be correct (Source: https://www.reddit.com/r/linux_gaming/comments/1sw83vn/i_added_12vhpwr12v2x6_power_connector_monitoring/):

> P.P.S.: Code was written using Claude Sonnet 4.6, I'm not fluent in Rust at all.

It was stated the work originated from https://github.com/Timic3/astral-power-monitoring, which also lead to https://github.com/jan-provaznik/sus and https://github.com/LibreHardwareMonitor/LibreHardwareMonitor/pull/2168. After also going through the related codebases, I think Linux hwmon/lm_sensors is the better place to upstream these efforts. The original author @Timic3 stated "NvAPI is not supported on linux, but there is a better way to fetch data from the chip, because Linux kernel exposes some functions to communicate via I2C." in their repo.

PR #995, opened 2026-04-27
feat: add hold/release cookie mechanism for profile state management

e-tho:
Clients that need to temporarily activate a profile had no safe way to restore the previous state on cleanup, and concurrent clients would silently overwrite each other's profile choice.

Introduce HoldProfile/ReleaseProfile commands modelled after power-profiles-daemon's HoldProfile/ReleaseProfile API. Acquiring a hold activates the requested profile and returns an opaque cookie; the previous state is saved automatically. Releasing the cookie restores it. Multiple concurrent holds are supported with last-write-wins semantics — releasing the active hold falls back to the next most recent one, and releasing the last hold restores the original state. If the client connection drops without releasing, the hold is released automatically.

### DIFF
--- a/lact-daemon/src/server.rs
+++ b/lact-daemon/src/server.rs
@@ -15,6 +15,7 @@ use std::fmt::Debug;
 use tokio::{
     io::{AsyncBufReadExt, AsyncRead, AsyncWrite, AsyncWriteExt, BufReader},
     net::{TcpListener, UnixListener},
+    sync::watch,
 };
 use tracing::{debug, error, info, instrument, trace};
 
@@ -116,6 +117,8 @@ pub async fn handle_stream<T: AsyncRead + AsyncWrite + Unpin>(
     stream: T,
     handler: Handler,
 ) -> anyhow::Result<()> {
+    let (_disconnect_tx, disconnect_rx) = watch::channel(false);
+
     let mut stream = BufReader::new(stream);
 
     let mut buf = String::new();
@@ -124,7 +127,7 @@ pub async fn handle_stream<T: AsyncRead + AsyncWrite + Unpin>(
 
         let maybe_request = serde_json::from_str(&buf);
         let response = match maybe_request {
-            Ok(request) => match handle_request(request, &handler).await {
+            Ok(request) => match handle_request(request, &handler, &disconnect_rx).await {
                 Ok(response) => response,
                 Err(error) => serde_json::to_vec(&Response::<()>::from(error))?,
             },
@@ -142,8 +145,12 @@ pub async fn handle_stream<T: AsyncRead + AsyncWrite + Unpin>(
     Ok(())
 }
 
-#[instrument(level = "debug", skip(handler))]
-async fn handle_request<'a>(request: Request<'a>, handler: &'a Handler) -> anyhow::Result<Vec<u8>> {
+#[instrument(level = "debug", skip(handler, disconnect_rx))]
+async fn handle_request<'a>(
+    request: Request<'a>,
+    handler: &'a Handler,
+    disconnect_rx: &watch::Receiver<bool>,
+) -> anyhow::Result<Vec<u8>> {
     match request {
         Request::Ping => ok_response(ping()),
         Request::SystemInfo => ok_response(system::info().await?),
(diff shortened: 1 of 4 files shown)


ilya-zlobintsev:
I understand the logic, but could you please explain on a bit higher level, what do you want to achieve and who will use this API? Is this to temporarily activate a profile and edit it, without making it be automatically applied at boot?
I'm not a big fan of having daemon logic bound to the fact that a specific client is connected.

e-tho:
I'm working on a modular system tuning daemon that activates during the lifetime of a process or on demand, coordinating backends like `logind`, `scx_loader`, `power-profiles-daemon`, and LACT. Basically a modernized GameMode.

Without a hold/release API, the only option is to manipulate LACT's global profile state directly: save the current profile, switch, disable auto-switching, restore on exit. It would break on unclean exit, conflict with LACT's profile watcher, and put restore responsibility on the client. Tying the hold to the connection solves this cleanly since LACT restores its previous state as soon as the client disconnects.

I believe that managing system state around a process is better handled by a dedicated orchestrator than built into each individual tool, and a single daemon that handles this cleanly for all backends seemed like the right solution. It's just a matter of days before I release the code publicly, if you'd prefer to take a look before making your decision.

ilya-zlobintsev:
This use case makes sense and I understand why you want to do this. 
The reason I'm somewhat reluctant about this PR is that the profile logic is already quite complex (especially around the profile watcher), and there have been concurrency bugs with it in the past such as leaking background watcher tasks. So I want to be extra careful specifically about holding extra state and stopping/starting the watcher on demand

e-tho:
I understand your concern given the history, but this implementation simply stops the watcher at the first hold and restarts it at the last release or on client disconnection, using the same `stop_profile_watcher` and `start_profile_watcher` calls that `set_profile` already uses. If there are concurrency issues around those, they would surface there too. If you identify any changes to make this safer, just let me know and I'll add them.

lact-schema/src/request.rs
```diff
@@ -85,6 +85,13 @@ pub enum Request<'a> {
         name: String,
         new_position: usize,
     },
+    HoldProfile {
+        name: String,
+        requester: String,
```
ilya-zlobintsev: This seems completely unused in the implementation except for logging purposes (which is useful, but still). I think it should at least be optional

lact-daemon/src/server.rs
```diff
@@ -116,6 +117,8 @@ pub async fn handle_stream<T: AsyncRead + AsyncWrite + Unpin>(
     stream: T,
     handler: Handler,
 ) -> anyhow::Result<()> {
+    let (_disconnect_tx, disconnect_rx) = watch::channel(false);
```
ilya-zlobintsev: This can be a basic [Notify](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html) rather than a watch channel, since we don't actually care about having a value, just the fact of notifying is enough to know that the client disconnected

lact-daemon/src/server/handler.rs
```diff
@@ -999,6 +1016,110 @@ impl<'a> Handler {
...
+    }
+
+    pub async fn release_profile(&self, cookie: u64) -> anyhow::Result<()> {
+        let was_top = {
+            let holds = self.profile_holds.borrow();
+            match holds.iter().rposition(|(c, _, _)| *c == cookie) {
+                Some(idx) => idx == holds.len() - 1,
+                None => bail!("Unknown profile hold cookie {cookie}"),
+            }
+        };
```
ilya-zlobintsev, line 1091: This could probably return the index so that you can use it to remove the hold, instead of calling `retain`

lact-daemon/src/server/handler.rs
```diff
@@ -999,6 +1016,110 @@ impl<'a> Handler {
...
+            self.stop_profile_watcher().await;
+            *self.profile_hold_snapshot.borrow_mut() = Some((snapshot_profile, snapshot_auto_switch));
+        }
+
+        let name_rc: Rc<str> = Rc::from(name.as_str());
+
+        let cookie = self.next_hold_cookie.get();
+        self.next_hold_cookie.set(cookie + 1);
+
+        let (drop_guard_tx, drop_guard_rx) = oneshot::channel::<()>();
```
ilya-zlobintsev: I get that the point of this channel is to automatically detect when the sender part is dropped, but this is not very clear behaviour,. I'd prefer to have this at least be explained in comments, or alternatively it can be an explicit message sent to the "holder" task, rather than relying on drop

ilya-zlobintsev (changes requested):
Looks fine logic-wise, at least I didn't manage to come up with cases where this breaks.

Also please use `cargo fmt` and `cargo clippy`

e-tho:
I removed the `requester` field entirely, would you prefer it optional?

ilya-zlobintsev:
> I removed the requester field entirely, would you prefer it optional?

It's fine as is.

ilya-zlobintsev (approved):

PR #997, opened 2026-04-30
feat: consolidate MessageDialogs under generic InfoDialog

makarov-roman:
Hi @ilya-zlobintsev,
tested everything besides dump_vbios

Here I want to create new InfoDialog which allows:
1. show generic info dialog
2. show error (with stacktrace in debug build)
3. show confirmation dialog (with optional timeout)

MessageDailogs are deprecated and will be removed in the next major gtk. So this is partially a preparation to v4.14 feature gate.

InfoDialog is not very generic as it describe different use cases, e.g. has presets for different messages
also it's **stateful** and allows only 1 instance of the dialog to be open (there was a similar check for 3 instances of error). 

also there are some misc changes:
- generate snapshot uses adw::Toast instead of MessageDialog
- use i18n for these dialogs


lact-gui/src/app/info_dialog.rs
```diff
@@ -0,0 +1,383 @@
...
+const RESPONSE_CONFIRM: &str = "confirm";
+
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+pub enum InfoDialogId {
+    Error,
+    EmbeddedDaemonInfo,
+    ResetConfigConfirmation,
+    SettingsConfirmation,
+    VersionMismatch,
+}
```
ilya-zlobintsev: Does not seem like the variants here are actually used for anything?
EDIT: nvm, I see that it is for the duplicate check

ilya-zlobintsev (changes requested):
UI-wise it's fine. but I don't really like the API/implementation. I think the callback mechanism is overkill for this, and kinda goes against the message passing architecture of relm.

Also, it looks like a lot of the code in the dialog is related to the countdown logic for confirmation, it would probably be fine to have a separate component just for the confirmation dialog, rather than making it all generic inside of one.

Additionally, have you seen `Alert` from `relm4-components`? It might be helpful: https://docs.rs/relm4-components/latest/relm4_components/alert/index.html

lact-gui/src/app/info_dialog.rs
makarov-roman: that's what I thought too, but already put to much effort to back down 😅 
upd: meant to response to the main message, not this thread

lact-gui/src/app.rs
```diff
@@ -1138,195 +1176,102 @@ impl AppModel {
...
+        widgets: &AppModelWidgets,
+    ) -> anyhow::Result<()> {
+        let path = self.daemon_client.generate_debug_snapshot().await?;
 
-                diag.run_async(|diag, _| {
-                    diag.hide();
-                })
+        let toast = adw::Toast::builder()
+            .title(format!("Debug snapshot saved at {path}"))
+            .button_label("Copy path")
```
ilya-zlobintsev, line 1259: Nitpick: maybe this could be an icon?

ilya-zlobintsev (approved):

lact-gui/src/app.rs
makarov-roman: @ilya-zlobintsev I like that it says copy path, otherwise they might assume that it copies the snapshot (which would be useful and better though)

PR #1008, opened 2026-05-05
fix: detect typed input in clock adjustment spinbuttons

aisk:
When type the value in the input, currently the apply button is still disabled. This PR fixed it.


ilya-zlobintsev (approved):
I'm a bit surprised that this works - normally the SpinButton sets the value of the underlying Adjustment, and the change event comes from that. I guess typing in a value (before pressing enter) does actually change the inner adjustment value, but without sending a signal through it.

But this does work, and it's a better UX, so LGTM

Issue #994, opened 2026-04-27
GPU Power State table UI inconsistencies


1nf3sT:
### Checklist

- [x] The problem is not listed in the [hardware support matrix](https://github.com/ilya-zlobintsev/LACT/wiki/Hardware-Support) as a known limitation. Examples of known driver-side issues are fans not turning on with a custom curve on RDNA3/4 and missing fan control on laptops.
- [x] I've included a [debug snapshot](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#reporting-issues) if the issue is related to hardware configuration
- [x] I've checked that the issue still happens in the latest [development build](https://github.com/ilya-zlobintsev/lact?tab=readme-ov-file#development-builds).

### Bug description

I believe I've found inconsistencies in the UI on both v8.5.0 and v9.0.0. 
The GPU Power State table comes and goes. I don't know if is related to the change in [PR #896](https://github.com/ilya-zlobintsev/LACT/pull/896). 
Please watch the two video clips of v8.5.0 and v9.0.0. 

[v9.0.0 - Power state table issues MP4](https://github.com/user-attachments/assets/5022d06b-acdb-4d23-b536-f3c11267c6fd)
[v8.5.0 - Power state table issues MP4](https://github.com/user-attachments/assets/17b2f4b5-77f5-430d-bfa6-6e7d0101317b)

[LACT-v0.9.0-snapshot-20260427-234305.tar.gz](https://github.com/user-attachments/files/27143188/LACT-v0.9.0-snapshot-20260427-234305.tar.gz)
[LACT-v0.8.5-snapshot-20260427-231441.tar.gz](https://github.com/user-attachments/files/27143187/LACT-v0.8.5-snapshot-20260427-231441.tar.gz)

### System info

```markdown
- LACT version:v8.5.0 & 9.0
- GPU model: RX 7900 XTX
- Kernel version: Linux 7.0.1-1-cachyos
- Distribution: CachyOS x86_64
```

makarov-roman:
there seems to be multiple problems:
- one core p-state is not reported at all -150hz for idle (it matches the snapshot we have for 7900xt)
- some data frames do not have active p-state data
- and, as you mentioned, visibility of core p-states column is not consistent

first 2 might be (or might be not) a driver issue. 

I'll look into 3rd point.

makarov-roman:
@1nf3sT can you help me to reproduce it?
here is a new test build with additional log, which logs p-state init + updates
[...]

PR #1019, opened 2026-05-12
chore: improve p-state reads logs

makarov-roman:
ref #994 

currently p-state read/parse errors are silenced. This change will add error logs with debug enabled and dump p-states readings with trace enabled.

### DIFF
--- a/lact-daemon/src/server/gpu_controller/amd.rs
+++ b/lact-daemon/src/server/gpu_controller/amd.rs
@@ -10,7 +10,7 @@ use crate::server::{
 use amdgpu_sysfs::{
     error::Error,
     gpu_handle::{
-        CommitHandle, GpuHandle, PerformanceLevel, PowerLevelKind, PowerLevels,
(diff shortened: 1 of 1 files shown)


lact-daemon/src/server/gpu_controller/amd.rs
```diff
@@ -1016,21 +1018,27 @@ impl GpuController for AmdGpuController {
...
-            memory_power_state: self
-                .handle
-                .get_memory_clock_levels()
-                .ok()
-                .and_then(|levels| levels.active),
-            pcie_power_state: self
-                .handle
-                .get_pcie_clock_levels()
-                .ok()
-                .and_then(|levels| levels.active),
```
ilya-zlobintsev, line 1033: It would be shorter to do an `.inspect_err` here instead of full match arms

ilya-zlobintsev (changes requested):

ilya-zlobintsev (approved):

PR #1023, opened 2026-05-14
feat: optimize gui bootstrap

makarov-roman:
before the change there were 3 reloads during bootstrap.
The old bootstrap path had:
- `ReloadProfiles`
- `GpuSelected`
- profile selector emitting `ReloadData { full: false }`
- then `ReloadData { full: true }`
- then another `ReloadData { full: false }

I also slightly refactored set_profile_data, by removing reload side effect

### DIFF
--- a/lact-gui/src/app.rs
+++ b/lact-gui/src/app.rs
@@ -534,8 +534,6 @@ impl AsyncComponent for AppModel {
             if let Err(err) = model.update_gpu_data_full(gpu_id, sender.clone()).await {
                 sender.input(AppMsg::Error(Arc::new(err)));
             }
-        } else {
-            warn!("no GPUs found during startup");
         }
 
         let widgets = view_output!();
--- a/lact-gui/src/app.rs
+++ b/lact-gui/src/app.rs
@@ -648,6 +648,18 @@ impl AppModel {
                 self.reload_profiles(state_sender).await?;
                 sender.input(AppMsg::ReloadData { full: false });
             }
+            AppMsg::ProfilesPolled(profiles) => {
+                let profiles_changed = self
+                    .profile_selector
+                    .model()
+                    .profiles_info_changed(profiles.as_ref());
+                self.profile_selector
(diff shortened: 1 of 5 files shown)


lact-gui/src/app.rs
```diff
@@ -488,14 +516,23 @@ impl AsyncComponent for AppModel {
...
+        if let Err(err) = model.reload_profiles(None).await {
+            sender.input(AppMsg::Error(Arc::new(err)));
+        }
+
+        if let Some(gpu_id) = initial_gpu_id {
+            if let Err(err) = model.update_gpu_data_full(gpu_id, sender.clone()).await {
+                sender.input(AppMsg::Error(Arc::new(err)));
+            }
+        }
+
```
ilya-zlobintsev, line 535: What's the advantage of calling these directly, rather than sending a message?

lact-gui/src/app/profiles.rs
```diff
@@ -22,10 +22,11 @@ use relm4::css;
 use relm4::prelude::DynamicIndex;
 use relm4::prelude::FactoryVecDeque;
 use relm4::{ComponentParts, ComponentSender};
+use std::sync::Arc;
 use tracing::debug;
 
 pub struct ProfileSelector {
-    profiles_info: ProfilesInfo,
+    profiles_info: Arc<ProfilesInfo>,
```
ilya-zlobintsev, line 29: What's the reason for this now being in an Arc? It doesn't seem like we clone it anywhere

lact-gui/src/app.rs
makarov-roman: That we can await for them

ilya-zlobintsev: Ah, makes sense, we want init to only be done after everything is loaded

lact-gui/src/app/profiles.rs
ilya-zlobintsev: Nevermind, I saw that it's for making `AppMsg` cloneable

ilya-zlobintsev (approved):

PR #1027, opened 2026-05-18
fix: fix regression with dirty init state

makarov-roman:
It also feels a bit as a workaround, but at least it's localized to this component, and not a global reload.


ilya-zlobintsev:
I've pushed a slightly different workaround, where instead of blocking, the signals get created only after UI is initialized.

PR #1028, opened 2026-05-19
feat: rework app menu using actions; add keyboard shortcuts for preferences and quit

GabMus:
As per the title. Here's a comparison between the old implementation using buttons and the new one:

| master | rework-menu |
|---|---|
|  |  |

Other notable changes include:

- Added tooltip to the menu button
- "Quit" action, mostly to bind it to Ctrl+Q as in most other apps
- "Preferences" action bound to Ctrl+comma, again conforming with most other apps
- The dump_vbios_action is in the application state, that's because it's enabled/disabled when gathering the device flags, so I need a way to keep a reference to it around, and this was probably the simplest way
- The application variable contains a reference to adw::Application (subclass of gtk::Application and I believe gio::Application as well), this is used to call its quit method when handling the newly introduced quit message. If you don't think this approach is good enough it's also possible to call root.application(), but that returns an option that needs to be unwrapped somehow. It's safe to do it in the context where it's ran, but I'd rather have it done just once.

### DIFF
--- a/lact-gui/src/app.rs
+++ b/lact-gui/src/app.rs
@@ -62,9 +62,11 @@ use pages::{
 use relm4::{
     AsyncComponentSender, Component, ComponentController, MessageBroker, RelmObjectExt,
     RelmWidgetExt,
+    actions::{AccelsPlus, ActionGroupName, RelmAction, RelmActionGroup},
     binding::BoolBinding,
     css,
     loading_widgets::LoadingWidgets,
+    new_action_group, new_stateless_action,
     prelude::{AsyncComponent, AsyncComponentParts},
     tokio::{self, time::sleep},
     view,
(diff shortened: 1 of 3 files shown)


lact-gui/src/app.rs
```diff
@@ -500,6 +463,46 @@ impl AsyncComponent for AppModel {
...
+        }));
+        application.set_accelerators_for_action::<PreferencesAction>(&["<Control>comma"]);
+        actions.add_action(RelmAction::<AboutAction>::new_stateless(move |_| {
+            APP_BROKER.send(AppMsg::ShowAboutDialog);
+        }));
+        actions.add_action(RelmAction::<QuitAction>::new_stateless(move |_| {
+            APP_BROKER.send(AppMsg::Quit);
+        }));
+        application.set_accelerators_for_action::<QuitAction>(&["<Control>q"]);
+        root.insert_action_group(AppActionGroup::NAME, Some(&actions.into_action_group()));
```
ilya-zlobintsev, line 500: It doesn't have to be changed in this PR, but this actions initialization is quite verbose. It would be nice if it could be shortened with a function or macro. This is how it was done before for example: https://github.com/ilya-zlobintsev/LACT/blob/v0.8.1/lact-gui/src/app.rs#L1122

ilya-zlobintsev (commented):
This actually used to be a `menu!`/actions thing in the past already, but it was removed when localization support was added, as at the time I thought the macro only supports literal strings. Nice to see that's not the case.

lact-gui/src/app.rs
GabMus: Added a macro as per your suggestion, let me know what you think

ilya-zlobintsev (approved):
Thanks! This is very concise now

PR #1031, opened 2026-05-20
feat: add custom loader

makarov-roman:
https://docs.gtk.org/gtk4/class.Svg.html would be certainly nicer, but imo looks good enough already :)
250ms min loader time is a common practice to not overload user with too many shifts in UI


ilya-zlobintsev:
Oops, I accidentally left review comments directly on the commit, rather than the PR: https://github.com/makarov-roman/LACT/commit/78ce0c16dff6809d4054aa266747392db65220b8

makarov-roman:
@ilya-zlobintsev updated

ilya-zlobintsev (approved):

Issue #996, opened 2026-04-28
Enable batch selection/adjustment of V/F curve points


FlightSuitRiot:
Can you enable the user to select multiple V/F curve points at once and adjust them all simultaneously in the V/F curve editor? Currently users can adjust one point at a time. Allowing batch selection and adjustment is helpful for undervolting because it allows the user to set a consistent offset along the V/F curve up to the desired voltage or clock maximum, and can also flatten the V/F curve after the maximum desired voltage/clock point. This feature is already in LACT with the "locked clocks" settings, it would just be translating it to the curve editor. It is similar to MSI Afterburner's V/F curve editor controls.

Ex: user has a 4070. Current curve has a V/F point at 950mv and 2445MHz. In V/F curve editor, if the user holds the "Shift" key and drags any point on the curve up or down, it applies the offset to ALL points along the curve. 

If the user holds the shift key and drags the cursor left or right anywhere on the graph, it soft-selects ALL points within the selected area (i.e., user holds shift and drags between 925mv and 1000mv, curve editor soft-selects all points in that range), and the user can then move just those selected points up and down in unison.

<img width="1291" height="790" alt="Image" src="https://github.com/user-attachments/assets/26d9c632-659b-4f1d-a516-eede0b6a36df" />

<img width="1227" height="788" alt="Image" src="https://github.com/user-attachments/assets/c2a38b24-5366-41b8-a1b9-bc966e867cae" />

ilya-zlobintsev:
You can already drag the whole curve by holding shift, and flatten it from the right click menu. You indeed can't move part of the curve though, at least without awkward "visible range" changes. 
Do you have something specific you want to do with the curve that can't be achieved with the current functionality, other than dragging a few (but not all) points at a time? I would rather cover an end goal of doing something with the curve in a simple and discoverable way, rather than just copying afterburner, because IMO afterburner's UX here is terrible (you have to aim for points that are a few pixels wide, the range selection is janky, and the modifier keys are not really intuitive)

lextra2:
@ilya-zlobintsev 
How about something like this (Quick ugly mockup)

<img width="1150" height="750" alt="Image" src="https://github.com/user-attachments/assets/5809826c-78bb-4e5b-90fb-212d6dfdfcad" />

Three fields where the user can enter some values:

**Min Voltage**: Start of the custom curve.

**Max Voltage**: Last point in the custom curve.
[...]

PR #1032, opened 2026-05-21
Add range selection feature for VF curve editor

sirno:
This implements basic batch selection as mentioned in #996:

- Points can be highlighted by dragging without touching nodes. 
- All selected points are adjusted by moving any of the selected points. 
- Dragging a point that is not selected will remove the selection.
- Clicking removes the selection, but its not cleanly implemented

There are still a few details until this is ready to merge:

- Some visual guide during selection, such as vertical bars with a shaded area
- Text showing position on the graph while dragging
- Immediate visual feedback when selecting outside the range
- Review choice of point style (the size multiplies twice now)

### DIFF
--- a/lact-gui/Cargo.toml
+++ b/lact-gui/Cargo.toml
@@ -37,6 +37,7 @@ i18n-embed-fl = { workspace = true }
 rust-embed = { workspace = true }
 
 plotters = { version = "0.3.7", default-features = false, features = [
+    "area_series",
     "line_series",
     "full_palette",
     "point_series",
--- a/lact-gui/src/app/pages/oc_page/vf_curve.rs
+++ b/lact-gui/src/app/pages/oc_page/vf_curve.rs
@@ -337,9 +337,21 @@ impl relm4::Component for VfCurveEditor {
             VfCurveEditorMsg::DragEnd => {
                 if self.dragging_point.take().is_some() {
                     APP_BROKER.send(AppMsg::SettingsChanged);
-                } else if self.allow_editing.value() && self.selected_range_start.get().is_some() {
-                    self.selected_range_end
-                        .set(self.hovered_coords.get().map(|(x, _)| x as usize));
+                } else if self.allow_editing.value()
+                    && let Some(selected_start) = self.selected_range_start.get()
+                {
+                    // let selected_end = self.hovered_coords.get().map(|(x, _)| x as usize);
+                    if let Some(selected_end) = self.hovered_coords.get().map(|(x, _)| x as usize)
+                        && self.points.borrow().iter().any(|p| {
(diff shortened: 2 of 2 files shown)


ilya-zlobintsev:
This is very cool! Nice that it also didn't need any extra input handling, and the current drag messages are enough

Some suggestions:
> Some visual guide during selection, such as vertical bars with a shaded area

Should be possible to use an plotters' [AreaSeries](https://docs.rs/plotters/latest/plotters/series/struct.AreaSeries.html) for this

> Text showing position on the graph while dragging

Would this just show cursor position, or the range which is selected?

> Review choice of point style (the size multiplies twice now)

I think maybe there will be no need to change the point size, if there's a background area + colour to differentiate them already.

Also, another idea: it might be nice to set the cursor style here when dragging, [this is done in the fan curve already](https://github.com/ilya-zlobintsev/LACT/blob/master/lact-gui/src/app/pages/thermals_page/fan_curve_frame.rs#L143).

sirno:
Cool, I am also quiet happy how neatly this fits in. I've added shading now and adjusted the cursor during selection. With both of these changes the selection feels quite good. I don't think the text indicator is really needed after all, because it is still possible to hover the points while dragging.

Can you give it a spin and see if you want anything else changed?

ilya-zlobintsev (changes requested):
I like it in terms of look and behaviour, I don't believe anything needs changes in that regard. (Also this is hopefully much more intuitive than afterburner)

Just one small issue that I've found: if you select an area, then disable curve editing, it ends up in a  weird state where you can still partially move the selection around. It should automatically unselect when disabling editing.

sirno:
The endpoint in the selection could still be moved. I've now frozen it, when editing is disabled. I think it may be helpful to make sure nothing about the visual state changes (including the selection) when editing is disabled. This is how it runs now.

Alternatively, it could also make sense to allow creating a selection while editing is disabled and only restrict editing of the actual vf points.

ilya-zlobintsev:
> The endpoint in the selection could still be moved. I've now frozen it, when editing is disabled. I think it may be helpful to make sure nothing about the visual state changes (including the selection) when editing is disabled. This is how it runs now.
> 
> Alternatively, it could also make sense to allow creating a selection while editing is disabled and only restrict editing of the actual vf points.

I don't think selection should be allowed when editing is disabled, there's nothing you can do with the selection anyway, so it is pointless. I also think it would be better if the selection was automatically removed when editing is disabled, or when "reset" is pressed.

Another small UX request: it would be nice if the selection was cancelled when the drag is released, but no points are selected. It's currently possible to do this kind of micro-select accidentally (including when just trying to click away from an existing selection).

sirno:
Alright, I've changed both.

ilya-zlobintsev (approved):
Thanks!

PR #1047, opened 2026-05-28
refactor: improve gui folder structure

makarov-roman:
I was working on power page and felt like the structure becoming a bit chaotic. The obvious first change is to group and move reusable components.

technically it's not a new problem, and we can copy any structure from react projects, e.g. - https://github.com/alan2207/bulletproof-react/blob/master/docs/project-structure.md

I don't really have preference, as long as it's clear to where to look/put what. Wdyt?


ilya-zlobintsev (commented):
This seems fine, it helps a bit to differentiate reusable components vs just pages through the structure. At the same time, I think this is a marginal difference compared to the current state.

lact-gui/src/app/components/oc_adjustment.rs
```diff
@@ -1,3 +1,4 @@
+#[path = "oc_adjustment_imp.rs"]
```
ilya-zlobintsev: This I don't like. We shouldn't override module paths unless there's a very good reason to, it is very unexpected compared to any other Rust module

makarov-roman:
> This seems fine, it helps a bit to differentiate reusable components vs just pages through the structure. At the same time, I think this is a marginal difference compared to the current state.

the idea was to make a bigger change, but when I tried, it didn't look as I expected, rolled back to the simpler version with features/ for global and top level features, and components/ for reusable components

ilya-zlobintsev (changes requested):
I don't like the `features` name to be honest, what was the problem with the old structure? I think for non-reusable components it makes sense to just keep them under `app`.

Also, I generally prefer this file structure:
```
thing.rs
thing/subthing.rs
```
Over
```
thing/mod.rs
thing/subthing.rs
```
Mostly because it is easier to find the right file when navigating using file search, as opposed to having to look for a file called `mod.rs`

makarov-roman:
well, as it happens when you do something for a long time, the way you did before feels right, and any other way feels wrong and confusing. It took me implementing these changes to understand that the problem I'm trying to solve doesn't really exist. Components folder still makes sense

ilya-zlobintsev (approved):

PR #1048, opened 2026-05-29
feat: add power page

makarov-roman:
Hi,
image

the problems I'm trying to solve:
1. OC page is overloaded
2. I want more screen real-estate to create MC charts with relevant to the context stats
3. PowerStates table looked a bit awkward with the expander

What changed:
1. new power page, which hosts p-states, limit, level + relevant stats
2. gpu_stats_section.rs is now a reusable component to display gpu stats
3. thermals page now uses the same component


makarov-roman:
image

lact-gui/src/app/pages/thermals_page.rs
```diff
@@ -102,34 +97,7 @@ impl relm4::Component for ThermalsPage {
...
-                    set_column_spacing: 10,
-                    set_homogeneous: true,
-                    set_min_children_per_line: 2,
-                    set_max_children_per_line: 4,
-                    set_selection_mode: gtk::SelectionMode::None,
-
-                    append_child = &InfoRow {
-                        set_name: fl!(I18N, "temperatures"),
-                        #[watch]
-                        set_value: model.temperatures.as_deref().unwrap_or("No sensors found"),
```
makarov-roman, line 134: this is a change in behaviour, was always rendered, while the main gpu-stats will hide it

lact-gui/src/app/components/gpu_stats_section.rs
```diff
@@ -36,7 +57,7 @@ pub enum GpuStatsSectionMsg {
 impl relm4::SimpleComponent for GpuStatsSection {
```
makarov-roman, line 60: probably we can use factory here, but I don't want to refactor the widget within this PR, unless it's required

makarov-roman:
the change is mostly structural. I have some ideas how to UI better, but I'd do it in the other PR to keep them smaller

PR #1055, opened 2026-06-07
fix: correct nvidia fan curve pwm to percent conversion

therummage:
The curve control loop converted PWM (0-255) to the percentage NVML expects with `/ 2.5`, mapping a 255 PWM to 102% - outside the valid 0-100 range. A curve top point of 1.0 was therefore rejected with INVALID_ARGUMENT, after which the daemon disabled fan control for the GPU. 

The divisor should be 2.55 - the inverse of the `* 2.55` percent-to-pwm conversion already used in this file (255 / 2.55 = 100).


ilya-zlobintsev (approved):

ilya-zlobintsev:
Thanks, not sure how this wasn't noticed until now

PR #1058, opened 2026-06-10
feat: merge software and hardware info pages into single information page

GabMus:
As per the title, this is what it looks like:

(disregard the "device not found" stuff at the bottom, my running daemon version is older and I can confirm the same happens on the master branch)

This accomplishes a few of things:

1. bring related information closer together, so that they can easily be inspected without having to jump back and forth between pages
2. remove one item from the sidebar, getting rid of some clutter

~~Number 2 in particular is interesting to me: I'm thinking of potentially getting rid of the sidebar entirely, moving to a more traditional and compact view switcher based navigation in the headerbar (that would be replaced by a bottom bar navigation on smaller window sizes as per the GNOME design guidelines).~~

~~This would mean moving the gpu and profile selects somewhere else (still thinking about where to put them), and the apply/revert buttons, which I think could fit nicely in a bottom revealer that appears/disappears using the same logic used currently for making them sensitive/insensitive.~~

See https://github.com/ilya-zlobintsev/LACT/pull/1058#issuecomment-4670830764

EDIT: I purposefully changed the i18n key for `info-page` to `information-page` to avoid collisions with existing translations, as `info-page` was key for "Software Info", which is different from a more general "Information" as it should be with this change.


makarov-roman:
> (disregard the "device not found" stuff at the bottom, my running daemon version is older and I can confirm the same happens on the master branch) 
you need optional deps - vulkan-tools/clinfo installed for the data to appear

I understand why you want to get rid of the sidebar, we just introduced it because horizontal navbar wasn't enough. 
Currently it's a but lacking, but there are multiple changes in progress which will use it - #1048 #1052 and there will be more. 

> I think could fit nicely in a bottom revealer that appears/disappears using the same logic used currently for making them sensitive/insensitive.

this was the case in version 0.8 and before. The reason we decided to change that - the reveler can be quite annoying because it either rendered on top of the UI elements you might be using, or you have to shift the whole page which doesn't feel nice

regarding merging software and hardware info.
imo, It makes sense only as an attempt to fit the current menu into the horizontal navbar, but since this is the goal I don't see why it should be done.

If it's merged - it's already borderline overloaded, and we might to expose more data there further.

GabMus:
I understand the design direction will still require the sidebar, this is perfectly fine as far as I'm concerned: if there is (or will be as it seems) a reason for it to be there then it's justified.

Regardless, I feel like merging the two pages still has value. As I already mentioned, it's reasonable to want to have a complete view of both hardware and software information back to back, without having to switch back and forth between pages.

As an aside, I feel like the layout of the "information sections" probably leaves a bit to be desired in terms of legibility. I haven't touched it yet as I haven't been able to think of a better way to display all of this information in a way that is more legible, but since we're already talking about it, if you have any ideas I'd be happy to put in the work to implement them.

Your intention of adding even more information to this section increases the magnitude of the problem, and I can see how my PR would be considered pejorative in that context, but I still think it's better than the lonely "Hardware Info" section that exists currently.

In summary, let's keep the discussion going, particularly again if you have any ideas on layout changes. There's probably a good compromise somewhere that I'm just not seeing.

ilya-zlobintsev (commented):
Personally, I'm not a huge fan of this. The UI is quite crowded with the change, and at the default window size, the OpenCL info doesn't fit:

Image

The info page was actually unified a long time ago, back when it was only hardware info + vulkan. Later, more things were added (namely OpenCL and more hardware entries), so it made sense to be split.

I agree that the page layout is not super great right now, but I don't think merging the two pages makes things much better. Most of the info is not really related to each other: you don't really care about hardware specifics when checking for vulkan feature support, or the lact version. 

And it will probably be even worse if more info is added or restructured (for example, it might be worth to split the AMD hardware block/cache information into a separate section, so it doesn't have to be behind a popup). With that, you'd end up with 5 sections on a single page.

GabMus:
Got it, it's probably for the best to drop this PR then and instead discuss about the information layout(s) more broadly

ilya-zlobintsev:
> Got it, it's probably for the best to drop this PR then and instead discuss about the information layout(s) more broadly

Agreed, I'm open to suggestions/contributions for layout improvements, I just don't think this specific change is it.

PR #1063, opened 2026-06-17
feat: add VRAM Fast Timing support for RDNA4 (gfx12) GPUs

l33tm4st3r:
- Add mem_fast_timing_active to ClocksInfo and mem_fast_timing to ClocksConfiguration in lact-schema.
- Read and write mem_fast_timing sysfs node in lact-daemon for AMD GPUs.
- Add "Fast Timing" gtk::Switch toggle under VRAM overclocking settings in the GUI.
- Update snapshots and test configs.


lact-daemon/src/server/gpu_controller/amd.rs
```diff
@@ -118,6 +118,26 @@ impl AmdGpuController {
         self.handle.hw_monitors.first().and_then(|mon| f(mon).ok())
     }
 
+    fn get_mem_fast_timing(&self) -> anyhow::Result<bool> {
+        let content = self.handle.read_file("gpu_od/mem_ctrl/mem_fast_timing")?;
```
ilya-zlobintsev, line 122: Where did you get this setting from? This does not exist in the kernel currently, is there a patch for it or something?

ilya-zlobintsev (changes requested):

lact-daemon/src/server/gpu_controller/amd.rs
l33tm4st3r: Hi @ilya-zlobintsev. You're completely right, my bad. I was messing around with some out-of-tree experimental patches for GFX12 from a staging tree, and I stupidly assumed it was closer to being mainlined/public than it actually is. 

Since this specific `sysfs` interface is definitely not upstream yet (and who knows if the final layout will even look like this), it makes no sense to have it in LACT right now. Apologies for jumping the gun with non-public paths, closing the PR!

PR #1083, opened 2026-06-22
fix: derive V/F curve bounds from offset limits

### LINKED ISSUE #1065: NVIDIA V/F curve frequency range not going high enough
### Checklist

- [x] The problem is not listed in the [hardware support matrix](https://github.com/ilya-zlobintsev/LACT/wiki/Hardware-Support) as a known limitation. Examples of known driver-side issues are fans not turning on with a custom curve on RDNA3/4 and missing fan control on laptops.
- [x] I've included a [debug snapshot](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#reporting-issues) if the issue is related to hardware configuration
- [x] I've checked that the issue still happens in the latest [development build](https://github.com/ilya-zlobintsev/lact?tab=readme-ov-file#development-builds).

### Bug description

Hi!
I'm trying to overclock/undervolt my RTX 3060 using the V/F curve editor, but I'm running into a funny little issue:
The graph isn't big enough!
What I mean by this, is that the graph seems to max out around 2100mhz, but my card seems to be able to go higher.
Sometimes the graph's Y axis (frequency)'s maximum value seems to jump a bit up depending on how the points are set but it's kind of unpredictable.
For example right now (screenshot attached below, ignore my curve being a bit "jumpy" at times) the maximum frequency I can set for a voltage is 2112mhz, but since that is not a "valid" frequency (card seems to accept frequencies in 7-8 (probably actually 7.5) mhz steps), in the end it applies 2107mhz which is the highest valid frequency at or below 2112

Image

(adjusting the "Visible range" sliders doesn't help)

Since 2107mhz seems to be stable for now, I want to try higher frequencies. (Unless the undocumented API doesn't allow higher?)

It would be nice if the graph went a little higher or something, or if this is a driver/hardware limitation, then it could somehow be made clearer.

Debug snapshot is included just in case.

[LACT-v0.9.1-snapshot-20260618-015521.tar.gz](https://github.com/user-attachments/files/29070846/LACT-v0.9.1-snapshot-20260618-015521.tar.gz)

### System info

```markdown
- LACT version: 0.9.0, 0.9.1 (current test release)
- GPU model: Nvidia GeForce RTX 3060 12gb (Palit Dual non-OC)
- Kernel version: 7.0.10-arch1-1
- Distribution: Arch Linux
```

makarov-roman:
fixes #1065
the idea is to utilize to offset limits to get theoretical extremes. This will squash the chart a bit, but it looks reasonable to me.

### DIFF
--- a/lact-gui/src/app/pages/oc_page/vf_curve.rs
+++ b/lact-gui/src/app/pages/oc_page/vf_curve.rs
@@ -32,6 +32,7 @@ use std::{cmp, fmt::Write as _};
 // In percentage
 const POINT_VOLTAGE_HOVER_MARGIN: f32 = 0.01;
 const POINT_FREQ_HOVER_MARGIN: f32 = 0.03;
+const MIN_VISIBLE_FREQ_RANGE_PADDING: u32 = 200;
 
 #[derive(Clone)]
 pub struct VfCurveEditor {
@@ -52,6 +53,7 @@ pub struct VfCurveEditor {
     hovered_point: Rc<Cell<Option<usize>>>,
     dragging_point: Rc<Cell<Option<usize>>>,
     drag_modifiers: Rc<Cell<gdk::ModifierType>>,
+    cached_y_spec: Rc<Cell<Option<(u32, u32)>>>,
 
     selected_range_start: Rc<Cell<Option<usize>>>,
     selected_range_end: Rc<Cell<Option<usize>>>,
@@ -284,6 +286,7 @@ impl relm4::Component for VfCurveEditor {
             hovered_point: Rc::new(Cell::new(None)),
             dragging_point: Rc::new(Cell::new(None)),
             drag_modifiers: Rc::new(Cell::new(gdk::ModifierType::empty())),
+            cached_y_spec: Rc::new(Cell::new(None)),
             selected_range_start: Rc::new(Cell::new(None)),
             selected_range_end: Rc::new(Cell::new(None)),
         };
@@ -317,7 +320,8 @@ impl relm4::Component for VfCurveEditor {
                         .gpu_offsets
                         .get(&0)
                         .map(|offset| (offset.min, offset.max));
-                    self.freq_range.set(Self::freq_range(&points, offset_range));
+                    self.freq_range
+                        .set(Self::freq_limits_range(&points, offset_range));
                 }
 
                 if points.is_empty() {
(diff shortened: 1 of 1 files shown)


ilya-zlobintsev (commented):
I'm thinking this can instead be added as a slider/spinbutton at the bottom, or even a checkmark to extend the range. The OC in the linked issue is quite extreme, 95% of users won't have offset values this high, and having the curve squished makes it more difficult to make fine adjustments

lact-gui/src/app/pages/oc_page/vf_curve.rs
```diff
@@ -747,6 +778,47 @@ impl VfCurveEditor {
...
+
+    fn visible_freq_range(points: &[NvidiaVfPoint], freq_range: (u32, u32)) -> Option<(u32, u32)> {
+        let min_freq = points
+            .iter()
+            .flat_map(|point| [point.freq, point.base_freq])
+            .min()?;
+        let max_freq = points
+            .iter()
+            .flat_map(|point| [point.freq, point.base_freq])
+            .max()?;
```
ilya-zlobintsev, line 808: It's probably safe to assume that you can take the first and last points for the frequency ranges. Technically it could be slightly wrong if someone makes a curve with a spike in the middle, but I don't think that's a problem

lact-gui/src/app/pages/oc_page/vf_curve.rs
```diff
@@ -51,6 +53,7 @@ pub struct VfCurveEditor {
     hovered_point: Rc<Cell<Option<usize>>>,
     dragging_point: Rc<Cell<Option<usize>>>,
     drag_modifiers: Rc<Cell<gdk::ModifierType>>,
+    cached_y_spec: Rc<Cell<Option<(u32, u32)>>>,
```
ilya-zlobintsev, line 57: I'm not sure if caching this is worth, it should be a simple lookup based on the start/end of the curve, which is cheap enough to perform per frame.
(I'd like to avoid having even more state in the editor if possible)

ilya-zlobintsev (changes requested):
I like the behaviour, just some nitpicks about the implementation

lact-gui/src/app/pages/oc_page/vf_curve.rs
ilya-zlobintsev: I tried changing this and it doesn't really work well, let's keep it as is

lact-gui/src/app/pages/oc_page/vf_curve.rs
ilya-zlobintsev: I tried implementing this, and realized the suggestion doesn't make sense, since we need to store the last pre-drag value

ilya-zlobintsev (approved):

PR #1086, opened 2026-06-23
chore: de-couple nvidia target temp from amd pmfw target temp

### LINKED ISSUE #895: Support Nvidia GPU Target Temperature
nvidia-smi expose this parameter to set up a custom target temperature for the GPU sensor.

```
    -gtt  --gpu-target-temp=    Set GPU Target Temperature for a GPU in degree celsius.
                                Requires administrator privileges

```

makarov-roman:
closes #895 

technically - no rush to merge it before the release.

I  tested config migration, apply/reset behaviour.

I'm not a big fan how this powerLimit like block looks, but it is what it is.

### DIFF
--- a/lact-gui/src/app/pages/thermals_page.rs
+++ b/lact-gui/src/app/pages/thermals_page.rs
@@ -13,6 +13,7 @@ use crate::{
     },
 };
 use amdgpu_sysfs::gpu_handle::fan_control::FanInfo;
+use anyhow::anyhow;
 use fan_curve_frame::{
     CurveSetupMsg, DEFAULT_SPEED_RANGE, DEFAULT_TEMP_RANGE, FanCurveFrame, FanCurveFrameMsg,
 };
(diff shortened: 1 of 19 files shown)


lact-gui/src/app/pages/thermals_page.rs
```diff
@@ -132,6 +145,41 @@ impl relm4::Component for ThermalsPage {
                 }
             },
 
+            PageSection::new(&fl!(I18N, "general-section")) {
```
ilya-zlobintsev: I don't like "General" as the name of the section. Maybe something like "Limits" would be better?

ilya-zlobintsev (changes requested):
This is fine UI-wise, but I would like to avoid adding the dedicated IPC method and `GpuController` trait function. I understand it's modeled after the PMFW API, but the reason it is that way is because:
- The unified `set_gpu_config` API that's used by the UI didn't exist at the time
- There is a dedicated "reset" command in sysfs

I think it should be possible to handle this just by setting the config field to None on "reset", and treating that as "set default value" in the daemon.

PR #1087, opened 2026-06-23
feat: expand monitor info

makarov-roman:
I was checking what else relevant can we expose from edid. hdr looks interesting, but I wouldn't dump whole struct inline. 

these 2 are no-brainers for both parsing and showing.

### DIFF
--- a/lact-daemon/src/server/display.rs
+++ b/lact-daemon/src/server/display.rs
@@ -1,5 +1,7 @@
 use anyhow::Context;
-use lact_schema::{DisplayConnector, DisplayInfo, DisplayManufactureDate, DisplaysInfo};
+use lact_schema::{
+    DisplayConnector, DisplayInfo, DisplayManufactureDate, DisplayRefreshRateRange, DisplaysInfo,
+};
 use std::{collections::BTreeMap, fs, path::Path};
 use tracing::warn;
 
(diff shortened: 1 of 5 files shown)


lact-daemon/src/server/display.rs
```diff
@@ -114,10 +116,30 @@ fn get_display_entry(path: &Path) -> anyhow::Result<(String, DisplayInfo)> {
...
+    edid.display_descriptors()
+        .iter()
+        .find_map(libdisplay_info::edid::DisplayDescriptorRef::range_limits)
+        .and_then(|range| {
+            (range.min_vert_rate_hz > 0 && range.max_vert_rate_hz >= range.min_vert_rate_hz)
+                .then_some(DisplayRefreshRateRange {
+                    min_hz: range.min_vert_rate_hz.try_into().ok()?,
+                    max_hz: range.max_vert_rate_hz.try_into().ok()?,
+                })
+        })
```
ilya-zlobintsev, line 140: I'm not sure if this is an entirely correct thing to report, it lists 56-76hz range for my monitor, which only ever runs at 60hz. Perhaps this can be parsed from the modes? (And then the dynamic one can be used for VRR)

Image

makarov-roman: @ilya-zlobintsev it's time for the monitor overclock :sunglasses: 
to me it reports effective vrr range.
checking

makarov-roman: @ilya-zlobintsev well, EDID is less useful than I initially though. both refresh range and bit depth are just "declarations" and can differ from properties of actual active link.
I'll close this PR because exposing these values can be misleading

ilya-zlobintsev: yeah, I've read in compositor implementations etc that EDID often lies :disappointed:

PR #1090, opened 2026-06-25
fix: nvidia displayport2 bandwidth calculation

makarov-roman:
@ilya-zlobintsev I was learning how you implemented dp2 bandwidth and noticed a potential problem.
What I found interesting - the are 2 values exposed on nvidia link config params, one for dp1 and one for dp2, but we're using only dp1

header says they're mutually exclusive - https://github.com/ilya-zlobintsev/LACT/blob/36e273fc90726b1db54faf5bcc61f5f9bc69f81e/lact-daemon/include/nvidia/src/common/sdk/nvidia/inc/ctrl/ctrl0073/ctrl0073dp.h#L1308

then I asked llm to check this path e2e, and it pointed to another possible problem - 

there are some valid values defined for nvidia which are below UHBR_RATE_THRESHOLD
https://github.com/ilya-zlobintsev/LACT/blob/36e273fc90726b1db54faf5bcc61f5f9bc69f81e/lact-daemon/include/nvidia/src/common/sdk/nvidia/inc/ctrl/ctrl0073/ctrl0073dp.h#L1348

the "fix" is rather ugly and mostly to demonstrate the problem. I have an Idea how to express it better, but want to get your feedback before, maybe I'm missing something

### DIFF
--- a/lact-daemon/src/server/gpu_controller/nvidia.rs
+++ b/lact-daemon/src/server/gpu_controller/nvidia.rs
@@ -1343,13 +1343,17 @@ impl GpuController for NvidiaGpuController {
                             match handle.get_dp_link_config(display_id) {
                                 Ok(params) => {
                                     *lanes = Some(params.laneCount.try_into()?);
-                                    *bandwidth = Some(if params.linkBW != 0 {
-                                        crate::server::display::dp1_rate_to_bandwidth(params.linkBW)
-                                    } else {
-                                        crate::server::display::dp2_rate_to_bandwidth(
+                                    *bandwidth = if params.linkBW != 0 {
+                                        Some(crate::server::display::dp1_rate_to_bandwidth(
+                                            params.linkBW,
+                                        ))
+                                    } else if params.dp2LinkBW != 0 {
+                                        Some(crate::server::display::dp2_rate_to_bandwidth(
                                             params.dp2LinkBW,
-                                        )
-                                    });
+                                        ))
+                                    } else {
+                                        None
+                                    };
                                 }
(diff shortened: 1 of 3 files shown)


lact-daemon/src/server/gpu_controller/nvidia.rs
```diff
@@ -1343,9 +1343,13 @@ impl GpuController for NvidiaGpuController {
...
-                                    *bandwidth = Some(
-                                        crate::server::display::dp_rate_to_bandwidth(params.linkBW),
-                                    );
+                                    *bandwidth = Some(if params.linkBW != 0 {
+                                        crate::server::display::dp_rate_to_bandwidth(params.linkBW)
+                                    } else {
+                                        use crate::server::display::UHBR_RATE_MULTIPLIER;
+
+                                        params.dp2LinkBW * UHBR_RATE_MULTIPLIER
+                                    });
```
ilya-zlobintsev: Can't we just pass whichever of the values is non-zero to `dp_rate_to_bandwidth`? The function handles both UHBR and non-UHBR values 
EDIT: disregard, I didn't finish reading the PR description

makarov-roman: @ilya-zlobintsev that's the second problem, it might not work for some defined <1000 UHBR values, not sure if they're used in real life, but at least they are defined in headers

makarov-roman:
okay, so my idea is to move vendor specific heuristics to their controllers and keep simple convertors function instead

ilya-zlobintsev (approved):

PR #1092, opened 2026-06-25
feat: support nvidia powermizer

makarov-roman:
seems like there is some problem on nvidia side. 
changes in powermizer are not reflected in Nvidia X Server app or nvidia-settings, but nvml readback reports correct value, also I see p-state shift when max performance is used.


lact-schema/src/config.rs
```diff
@@ -41,6 +42,7 @@ pub struct GpuConfig {
     pub pmfw_options: PmfwOptions,
     #[serde(default, skip_serializing_if = "NvidiaThermalOptions::is_empty")]
     pub nvidia_thermal_options: NvidiaThermalOptions,
+    pub power_mizer_mode: Option<PowerMizerMode>,
```
makarov-roman, line 45: @ilya-zlobintsev currently this is saved into config with PascalCase, which is not consistent with rest of the config. Do know any tricks to handle this? I can only imagine hacky workarounds, like create some dervid enum just for config - ConfPowerMizerMode, and define rename_all=snake_case there,.

ilya-zlobintsev: It's possible to have a `#[serde(with = "customthing")]` attribute to define custom serialization/deserialization for that field that can contain custom logic. But I don't think it matters much, having the field in PascalCase is not a big problem

Cargo.toml
```diff
@@ -14,6 +14,7 @@ edition = "2024"
 
 [workspace.dependencies]
 amdgpu-sysfs = { version = "0.21.0", features = ["serde"] }
+nvml-wrapper = { git = "https://github.com/Cldfire/nvml-wrapper.git", rev = "eb47417eede43443f139053479618e96ad32893d" }
```
ilya-zlobintsev: This URL seems old, where did you get it? It's https://github.com/rust-nvml/nvml-wrapper now

lact-daemon/src/server/gpu_controller/nvidia.rs
```diff
@@ -1013,32 +1085,10 @@ impl GpuController for NvidiaGpuController {
...
-                    && current_cap != default_cap
-                {
-                    debug!("resetting power cap to {default_cap}");
-                    device
-                        .set_power_management_limit(default_cap)
-                        .context("Could not reset power cap")?;
-                }
+            if let Some(mode) = config.power_mizer_mode {
+                apply_power_mizer_mode(&mut device, mode)?;
             }
```
ilya-zlobintsev: This has no reset path, it will not clean up the setting on daemon shutdown/config reset

ilya-zlobintsev (changes requested):
The UI seems to be very bugged: any configured setting is not shown (the selector is always on auto), and it also highlights the apply/reset buttons on app launch when it shouldn't (probably related to the first problem)

lact-daemon/src/server/gpu_controller/nvidia.rs
makarov-roman: I switch to the same behaviour as on AMD - apply Auto when no value is supplied

ilya-zlobintsev (approved):
Works well now, including reset on daemon shutdown

PR #1096, opened 2026-06-26
fix: blackwell nvapi temps parsing

makarov-roman:
Seems like temps mapping on blackwell is different from ada.

this PR takes mapping from
https://github.com/LibreHardwareMonitor/LibreHardwareMonitor/pull/1878
their temp1 matches 9, temp2 matches 10, e.t.c

I tested sensors on the server and 9 matches core temp, 10 seems like vram, others are invalid including 15. Haven't tested lact itself.

### DIFF
--- a/lact-daemon/src/server/gpu_controller/nvidia/nvapi.rs
+++ b/lact-daemon/src/server/gpu_controller/nvidia/nvapi.rs
@@ -320,15 +320,16 @@ impl NvApiThermals {
     }
 
     pub fn hotspot(&self, arch: Option<&DeviceArchitecture>) -> Option<i32> {
-        match arch {
-            Some(DeviceArchitecture::Blackwell) => None,
-            _ => self.get_value(9),
+        if arch.is_some_and(|arch| arch.as_c() >= DeviceArchitecture::Blackwell.as_c()) {
+            None
+        } else {
+            self.get_value(9)
         }
     }
 
-    pub fn vram(&self, arch: Option<&DeviceArchitecture>) -> Option<i32> {
-        match arch {
-            Some(DeviceArchitecture::Blackwell) => self.get_value(10),
+    pub fn vram(&self, vram_type: Option<&str>) -> Option<i32> {
+        match vram_type {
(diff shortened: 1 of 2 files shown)


lact-daemon/src/server/gpu_controller/nvidia/nvapi.rs
```diff
@@ -318,12 +319,18 @@ impl NvApiThermals {
...
 
-    pub fn vram(&self) -> Option<i32> {
-        self.get_value(15)
+    pub fn vram(&self, arch: Option<&DeviceArchitecture>) -> Option<i32> {
+        match arch {
+            Some(DeviceArchitecture::Blackwell) => self.get_value(10),
+            _ => self.get_value(15),
+        }
     }
 }
```
ilya-zlobintsev: I'm guessing this isn't actually based on the generation, but rather on the VRAM type. The majority of Blackwell is GDDR7, but there is at least one Blackwell GDDR6 card (RTX 5050). Ideally we'd test to confirm, but my hunch is that checking for GDDR7 here would be more correct

lact-daemon/src/server/gpu_controller/nvidia/nvapi.rs
```diff
@@ -318,12 +319,18 @@ impl NvApiThermals {
...
     }
 
-    pub fn hotspot(&self) -> Option<i32> {
-        self.get_value(9)
+    pub fn hotspot(&self, arch: Option<&DeviceArchitecture>) -> Option<i32> {
+        match arch {
+            Some(DeviceArchitecture::Blackwell) => None,
+            _ => self.get_value(9),
+        }
     }
```
ilya-zlobintsev, line 328: I'm not sure about this, I've seen hotspot temp be reported on Blackwell before, and it's different from the NVML-reported GPU temp. Or do you think the reading is not correct?

makarov-roman: it's always within 1 degree difference due to different rounding.

PR #1097, opened 2026-06-27
fix: add workaround for amd broken p-state indexes

makarov-roman:
they might fix it tomorrow, they might never fix it, so I think we should handle it


ilya-zlobintsev:
Since we have the sleep state in 9070xt test data, could we have the config apply tests cover this? Should be enough to add some pstate settings to `tests/snapshots/amd/rx9070xt/config.yaml`

makarov-roman:
> Since we have the sleep state in 9070xt test data, could we have the config apply tests cover this? Should be enough to add some pstate settings to `tests/snapshots/amd/rx9070xt/config.yaml`

I don't think we should couple a pure function unit test with the snapshot, it's a bit ambiguous. Probably nobody will touch this function ever, so I can remove the tests if you think they're too verbose.

ilya-zlobintsev:
What I meant, at `tests/snapshots/amd/rx9070xt/card1/device/pp_dpm_sclk` we currently have a snapshot containing the "proper" format:
```
S: 19Mhz *
0: 615Mhz
1: 800Mhz
2: 888Mhz
3: 1000Mhz
```
It would be nice if we had one with the current broken state, that exercises that a custom pstates config gets applied correctly anyway. We could put it into the 9070 non-xt data for example

Issue #1101, opened 2026-06-29
feature: add auto/ligh/dark light theme switch into preferences.


makarov-roman:
There are cases where  systems don't have a working system preference, or user might prefer certain look. 
this is how gnome builder did it:

<img width="316" height="362" alt="Image" src="https://github.com/user-attachments/assets/dbbdd2c8-7f39-4a9e-b03f-898c5d006885" />

makarov-roman:
Hi @GabMus, I had an impression that you'd like to work on UI features, let me know if you're interested to take this one.

GabMus:
Sure, I can try and give this one a jab

GabMus:
All the existing implementations of the color scheme switcher you provided a screenshot for are licensed as GPL3, while LACT is licensed as MIT, which I believe is incompatible. Any insight on how to navigate this?

makarov-roman:
@GabMus it doesn't have to be the same switch, a simple worded (auto/dark/light) toggle (or even select, because we cannot use adwaita toggle due to 1.4 feature gate) in preferences dialog is enough.

ilya-zlobintsev:
Implemented in #1103

PR #1103, opened 2026-06-30
feat(#1101): auto/force light/force dark color scheme preference

GabMus:
As per the title, only issue is light mode does not work with the built in breeze style

### DIFF
--- a/lact-gui/i18n/en/lact_gui.ftl
+++ b/lact-gui/i18n/en/lact_gui.ftl
@@ -295,6 +295,11 @@ ui = UI
 daemon = Daemon
 about = About
 
+color-scheme = Color Scheme
+color-scheme-auto = System
+color-scheme-light = Light
+color-scheme-dark = Dark
+
 # Crash page
 crash-page-title = Application Crashed
 exit = Exit
--- a/lact-gui/src/app.rs
+++ b/lact-gui/src/app.rs
@@ -365,6 +365,7 @@ impl AsyncComponent for AppModel {
         if let Err(err) = styles::apply_theme(CONFIG.read().theme) {
             error!("could not apply theme: {err:#}");
         }
+        CONFIG.read().color_scheme.apply();
 
         let (daemon_client, conn_err) = match args.tcp_address {
             Some(remote_addr) => {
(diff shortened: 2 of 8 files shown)


lact-gui/src/app.rs
```diff
@@ -488,6 +488,8 @@ impl AsyncComponent for AppModel {
         application.set_accelerators_for_action::<QuitAction>(&["<Control>q"]);
         root.insert_action_group(AppActionGroup::NAME, Some(&actions.into_action_group()));
 
+        CONFIG.read().color_scheme.apply();
```
ilya-zlobintsev: Could this be moved up, to be next to the `styles::apply_theme` call? It would make sense for the styling to be together

lact-gui/src/app.css
```diff
@@ -13,3 +13,46 @@
...
+  min-width: 12px;
+  min-height: 12px;
+  transform: translate(27px, 14px);
+  padding: 2px;
+}
+checkbutton.color-scheme-switcher-checkbutton radio:checked {
+  -gtk-icon-source: -gtk-icontheme("object-select-symbolic");
+  background-color: @theme_selected_bg_color;
+  color: @theme_selected_fg_color;
+}
```
ilya-zlobintsev: None of these styles seem to actually be referenced by the switcher?

I think they might not even be necessary, it's fine for the switcher to be basic.

lact-gui/i18n/en/lact_gui.ftl
```diff
@@ -295,6 +295,11 @@ ui = UI
 daemon = Daemon
 about = About
 
+color-scheme = Color Scheme
+color-scheme-auto = System Style
+color-scheme-light = Light Style
+color-scheme-dark = Dark Style
```
ilya-zlobintsev: Having "Style" on the button feels unnecessary here, just System/Light/Dark should be clear enough

ilya-zlobintsev (changes requested):

lact-gui/src/app.css
GabMus: oops, residue from previous impl sorry

GabMus:
should have solved everything

ilya-zlobintsev (approved):
Thanks

PR #1108, opened 2026-07-05
Fan control for Intel Xe

WizardlyBump17:
This Pull Request adds fan control for Intel Xe graphics


WizardlyBump17:
Hey @ilya-zlobintsev, I am currently working on the get_stats, adding the fan stuff and I am taking the AMD implementation as an example. I noticed that sometimes the hwmon is used and sometimes the GpuConfig is used. Which one should I use? Intel exposed everything through hwmon sysfs.

Intel also exposed control for all fans, 1-3, while LACT seems to think that all fans work together. How should I make things configurable? Right now I am checking for `pwm1` to see if there is fan control support and I am making changes to all fans

lact-daemon/src/server/gpu_controller/intel.rs
```diff
@@ -667,6 +677,63 @@ impl GpuController for IntelGpuController {
...
+                                self.write_hwmon_file(&[&pwm_enable], "1")
+                                    .context("Could not enable manual fan control")?; //if it is 0 or 2, you wont be able to control the RPM
+                                self.write_hwmon_file(&[&pwm], &pwm_value.to_string())
+                                    .context("Could not set fan RPM")?;
+                            }
+                        },
+                        lact_schema::FanControlMode::Curve => {
+                            if (settings.curve.0.len() != 10) {
+                                return Err(anyhow!("The fan curve needs exactly 10 points"));
+                            }
```
ilya-zlobintsev: Is this accurate? Can you not write a curve with fewer points?

ilya-zlobintsev (commented):
Some feedback:
- The fan numbers should be discovered based on what's there in the hwmon, not hardcoded to 3
- Please run `cargo fmt` and `cargo clippy` on the code
- It will be useful if the debug snapshot (e.g. at `tests/snapshots/intel/b580/card1/device/hwmon/hwmon4`) would include the new fan data

ilya-zlobintsev:
> I noticed that sometimes the hwmon is used and sometimes the GpuConfig is used. Which one should I use? Intel exposed everything through hwmon sysfs.

The config is used on AMD to tell if LACT has configured the fan curve or not, and which settings have been overridden. It is better to fetch from hwmon when possible, since that will have the accurate hardware state.

> while LACT seems to think that all fans work together. How should I make things configurable?

All fans should be configured together. Technically Nvidia could also expose fan-specific control, but it is not currently supported and I'm not sure about how useful that would be.

lact-daemon/src/server/gpu_controller/intel.rs
WizardlyBump17: There are only 10 exposed points. It looks like you can specify points with the same temperatures, so we could let the user send one point and assing the rest to the same speed and temperature, but we cant give them more than the exposed points. The patchwork says they must be in increasing order, so I am checking for that too (not commited yet)

ilya-zlobintsev: The auto-filling if there are not enough points might be a good idea, alternatively this needs to be handled so that the UI gets a 10-point curve by default

WizardlyBump17:
If the xe driver starts complaining about your curve, set the control mode to Automatic, apply, then Static 100%, apply, then configure your curve.

The xe driver will override your curve points to the static speed you choose, at least for the revision 1 of the patch

WizardlyBump17:
@ilya-zlobintsev it is done.

I still cant get a debug snapshot because the kernel keeps crashing. This looks like a kernel bug.

We have the usual fan settings: automatic, curve and static.

For the curve, we omit the points with the same speed, but only the last ones; example: 10 0.1, 20 0.15, 30 0.2, 40 0.2, 50 0.25, 60 0.3, 70 0.3, 80 0.3 -> 10 0.1, 20 0.15, 30 0.2, 40 0.2, 50 0.25, 60 0.3. If the user specifies an amount of points less than the exposed ones, we auto complete them with the same speed and 1ºC from each other; the driver rejects points with the same temperature or with less than 1ºC difference.

If all points are the same speed, we say it is static.

No errors on a kernel without fan control. When the user boots into a kernel with fan control for the first time, the xe driver already has a curve configured (a shitty one for people with bad airflow).

I am a Java developer, but not a Rust developer, so I carried my Java mannerisms. If there is something wrong or bad done, let me know

lact-daemon/src/server/gpu_controller/intel.rs
```diff
@@ -224,10 +226,10 @@ impl IntelGpuController {
                 let entries = fs::read_dir(hwmon_path).ok()?;
                 for entry in entries.flatten() {
                     if let Some(name) = entry.file_name().to_str()
-                        && let Some(infix) = name
+                        && let Some(_infix) = name
                             .strip_prefix(file_prefix)
                             .and_then(|name| name.strip_suffix(file_suffix))
-                        && !infix.contains('_')
+                    // && !infix.contains('_')
```
ilya-zlobintsev, line 232: Why was this changed? Note that this code is used on both 915 and xe, so there might be cases you don't see on your GPU

lact-daemon/src/server/gpu_controller/intel.rs
```diff
@@ -570,6 +572,239 @@ impl IntelGpuController {
...
+        return self.match_hwmon_files(&["pwm1"]).next().is_some();
+    }
+
+    fn get_hwmon_controllable_points_amount(&self) -> u8 {
+        self.read_hwmon_files::<String>("pwm1_", "_pwm").count() as u8
+    }
+
+    fn get_hwmon_fan_curve(&self) -> Option<FanCurve> {
+        if !self.has_fan_control()  {
+            return Option::None;
```
ilya-zlobintsev: `None` and `Some` are always in scope in rust, you don't need to write `Option::`

lact-daemon/src/server/gpu_controller/intel.rs
```diff
@@ -570,6 +572,239 @@ impl IntelGpuController {
...
+
+        for point in 1..=amount {
+            let pwm_file: String = format!("pwm1_auto_point{point}_pwm");
+            let temp_file: String = format!("pwm1_auto_point{point}_temp");
+
+            let pwm: i32 = self
+                .read_hwmon_file(&[&*pwm_file], false)
+                .map(|value: u64| value as i32)
+                .context("Could not read the curve")
+                .unwrap();
```
ilya-zlobintsev, line 611: Please do not use `.unwrap()` for such things. The reads could randomly fail for any reason, the error should be reasonably handled (and realistically, since this is called through `get_stats`, just log it)

ilya-zlobintsev: Applies to multiple places here

lact-daemon/src/server/gpu_controller/intel.rs
```diff
@@ -570,6 +572,239 @@ impl IntelGpuController {
...
+            let temp: f32 = self
+                .read_hwmon_file(&[&*temp_file], false)
+                .map(|value: u64| value as f32)
+                .context("Could not read the curve")
+                .unwrap();
+
+            points.push((temp as i32 / 1000, pwm as f32 / 255.0));
+
+            info!("{}", temp);
```
ilya-zlobintsev: Debugging leftover? This should not be logged, it will be very spammy

lact-daemon/src/server/gpu_controller/intel.rs
```diff
@@ -570,6 +572,239 @@ impl IntelGpuController {
...
+
+            let speed: f32 = point.1;
+
+            if last_speed != speed  {
+                break; //we might be in the middle of the curve and we are not that smart to figure out how to handle that
+            }
+
+            last_speed = speed;
+
+            points.remove(point_index as usize + 1); //if the speed is the same as the last, we probably auto completed it, so we dont need to show it here
```
ilya-zlobintsev: I don't think this is a safe assumption to make. You could make such a curve on your own, in which case reading it back won't make any sense. If you want to differentiate between reported and configured points, you can compare the curve with the current config (it is available in `get_stats`)

lact-daemon/src/server/gpu_controller/intel.rs
```diff
@@ -570,6 +572,239 @@ impl IntelGpuController {
...
+        for fan_index in 1..=fans {
+            let pwm_enable = format!("pwm{fan_index}_enable");
+            self.write_hwmon_file(&[&pwm_enable], "1")
+                .context("Could not enable manual fan control")?;
+        }
+
+        for (i, (temperature, speed)) in curve.0.iter().enumerate() {
+            let point = i + 1;
+            let pwm = (*speed * 255.0) as u8;
+            let temp = (f64::from(*temperature) * 1000.0) as i32;
```
ilya-zlobintsev: Is there a reason why this is not just `*temperature * 1000`?

lact-daemon/src/server/gpu_controller/intel.rs
```diff
@@ -570,6 +572,239 @@ impl IntelGpuController {
...
+                    )
+                    .context("Could not set the RPM for a point in the curve")?;
+                }
+            }
+        }
+
+        Result::Ok(true)
+    }
+
+    fn is_valid_fan_curve(&self, curve: FanCurve) -> (bool, String) {
```
ilya-zlobintsev: In rust, for error handling, you should return a `Result` (in this case, an `anyhow::Result` would be easiest), instead of this status bool and error

ilya-zlobintsev (changes requested):
Logic seems to be reasonable overall, except for the flat curve "trimming" that I commented below.

There are a few stylistic things also, please run `cargo clippy` and `cargo fmt`, it will cover most of them.

As for the kernel crashing on debug snapshot - that is really bad, and I'd like to figure it out, because it will make it impossible for Arc users to make proper bug reports. Does it happen on a mainline kernel without the fan control patches?

lact-daemon/src/server/gpu_controller/intel.rs
WizardlyBump17: When I started I tried to read `read_hwmon_files::<String>("pwm1_", "_pwm")` and I was wondering why it was empty, then I saw that the `read_hwmon_files` was ignoring the `_`, so I checked where the method was being used and I determined that it wouldnt hurt to remove that check. This method had only one use before: read the temperature, which works the same for xe and i915; I updated the code to read `_input` instead of `input`

lact-daemon/src/server/gpu_controller/intel.rs
WizardlyBump17: It was `(*temperature as f64 * 1000.0) as i32;`, but `cargo clippy` changed that

lact-daemon/src/server/gpu_controller/intel.rs
WizardlyBump17: Do you have any idea on how to compare?

lact-daemon/src/server/gpu_controller/intel.rs
WizardlyBump17: How would I do that?

lact-daemon/src/server/gpu_controller/intel.rs
ilya-zlobintsev: But why do you need to cast it to an `f64`? Both the input and output are `i32`

lact-daemon/src/server/gpu_controller/intel.rs
ilya-zlobintsev: You don't need to return a `String` here at all, it should just be a `Result<(), Error>`

lact-daemon/src/server/gpu_controller/intel.rs
ilya-zlobintsev: You can pass `gpu_config` into this function, read all of the curve points, and if a custom fan curve is configured, only keep the number of points that is set in the config.

ilya-zlobintsev:
@WizardlyBump17 I've did some cleanup of the code, could you check if it still works as expected?

Additionally, did you get a chance to look at the crash when taking a debug snapshot? Does it still happen with the latest version?

WizardlyBump17:
Thank you. Everything seems to work.

As for the crash, it still happens. I tested on the patched kernel and 7.1.5 and the crash happened on both. I will try 7.2.0 once it hits mainline. This is the dmesg from the crash:
[dmesg.202608162129.log](https://github.com/user-attachments/files/31125349/dmesg.202608162129.log)


I will add a new method `get_hwmon_fan_curve` that has no parameters and rename the current one so people in the future dont expect the current one to return the exact fan curve from hwmon

ilya-zlobintsev:
@WizardlyBump17 could you check if adding `"intel_pmt"` to `SNAPSHOT_EXCLUDED_FILENAME_PREFIXES` in `handler.rs` fixes the crash?

WizardlyBump17:
> @WizardlyBump17 could you check if adding `"intel_pmt"` to `SNAPSHOT_EXCLUDED_FILENAME_PREFIXES` in `handler.rs` fixes the crash?

Yep, that fixed it.

[LACT-v0.10.1-snapshot-20260817-201818.tar.gz](https://github.com/user-attachments/files/31153804/LACT-v0.10.1-snapshot-20260817-201818.tar.gz)

WizardlyBump17:
I didnt realize before, but I realized now. Your truncated curve is removing more points than it should

image


This is what it should be with the default curve:
image

I think the old logic of omitting the last points that have the same speed is better than yours.


With my code back (`- 2` instead of `- 1`)
image

ilya-zlobintsev:
> I think the old logic of omitting the last points that have the same speed is better than yours.

My concern with this is that it is possible to manually configure a curve that has repeating speeds at multiple temperatures, and that would get incorrectly truncated.

I've updated the logic to account for both - it now checks for repeating points, but removes only as long as you don't have them in the config. 
Please try it out, it should work correctly now.

lact-daemon/src/server/gpu_controller/intel.rs
```diff
@@ -624,6 +625,12 @@ impl IntelGpuController {
 
     fn get_truncated_hwmon_fan_curve(&self, config: Option<&GpuConfig>) -> Option<FanCurveMap> {
         let mut points: Vec<(i32, f32)> = self.get_hwmon_fan_curve_points()?;
+
+        if self.get_hwmon_fan_control_mode() == Some(Static) {
+            let single_point: [(i32, f32); 1] = [*points.first().unwrap()];
+            return Some(single_point.into_iter().collect());
+        }
```
ilya-zlobintsev, line 632: Please change the `points.first().unwrap()` to `points.first()?`, so that it returns `None` if `points` is empty. Otherwise this will crash (a very possible scenario - someone sets a static fan control, then boots into an old kernel that does not have fan control support, and now just fetching info crashes)

ilya-zlobintsev:
Looks like the kernel patches are still not merged, so in practice most people won't be able to use this.
However, to avoid generating conflicts I'll merge this now.

PR #1110, opened 2026-07-10
chore: update rust deps

makarov-roman:
multiple deps updates are needed for mc chart. We also have to finally handle gtk feature gate

### DIFF
--- a/lact-daemon/src/tests/mock_fs.rs
+++ b/lact-daemon/src/tests/mock_fs.rs
@@ -1,12 +1,9 @@
 use easy_fuser::{
-    FuseHandler,
-    templates::{
+    delegate_fs,
+    fuse_parallel::prelude::*,
+    fuse_presets::{
         DefaultFuseHandler,
         mirror_fs::{MirrorFsReadOnly, MirrorFsTrait},
-    },
-    types::{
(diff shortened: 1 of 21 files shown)


lact-daemon/src/tests/mock_fs.rs
```diff
@@ -16,27 +13,37 @@ use std::{
...
+
+    delegate_fs! { mirror_fs, [
+        flush, fsync, lseek, read, release, access, getattr, listxattr, lookup, open, readdir,
+        readlink
+    ] }
+
+    delegate_fs! { default_fs, [
+        copy_file_range, fallocate, create, mkdir, mknod, removexattr, rename, rmdir, symlink,
+        unlink, bmap, forget, fsyncdir, getlk, ioctl, link, opendir, releasedir, setlk, statfs
+    ] }
```
ilya-zlobintsev, line 42: I don't like this API they're doing, but oh well.. maybe using `fuser` directly would be easier than `easy_fuser` (lol)

ilya-zlobintsev (approved):
I also added a fix for this warning, it was quite annoying to be getting it on every build
```
warning: the following packages contain code that will be rejected by a future version of Rust: proc-macro-error2 v2.0.1
```

PR #1115, opened 2026-07-16
chore: bump gtk feature gate to v4_14

makarov-roman:
this seems to be the last blocker for mc charts


lact-gui/src/app/utils/css_colors.rs
```diff
@@ -0,0 +1,123 @@
...
+
+            #[name = "theme_text_color"]
+            gtk::Label {
+                add_css_class: "color-probe-theme-text-color",
+            },
+
+            #[name = "borders"]
+            gtk::Label {
+                add_css_class: "color-probe-borders",
+            },
```
ilya-zlobintsev: I can't say I'm a fan of this approach with fake widgets. If there is no better way to query colours, I think we can keep the current StyleManager usage with `#[allow(deprecated)]`. Deprecations won't be removed until GTK5, so it's unlikely to cause problems in practice

At some point it might be also acceptable to just have our own list of colours for the chart separately for dark/light modes, and the only thing queried from the system theme would be the accent colour.

ilya-zlobintsev (changes requested):

lact-gui/src/app/utils/css_colors.rs
makarov-roman: Why though? Do you see potential issues? I think it's intended approach with this new Widget::color() api (I read about it on their discourse a while ago, can try to find the link).

ilya-zlobintsev: Probably no behavoiur issues, but having invisible widgets just for the purpose of probing their colours feels quite hacky to me. The main reason the colour discovery was made were custom GTK themes, which is no longer a big problem with libadwaita and the custom breeze version.

makarov-roman: There are themes for adwaita, so it's still relevant. The second use case - dark/light switch, which can probably derived from the app state.

I don't think `#[allow(deprecated)]` is a good idea, because we'd have to make the same decision later.

my take - we should either go with probe or remove css var reading in favor of static presets for light/dark charts.

ilya-zlobintsev: Let's do the static presets then. I'm concerned about the probe widgets affecting things like focus/input handling/layout in non obvious ways

ilya-zlobintsev (approved):

PR #1118, opened 2026-07-18
feat: add configurable NVIDIA initialization mode

hurryman2212:
Problem: If NVML initialization fails while no NVIDIA GPU is available, LACT's existing re-enumeration logic cannot create an NVIDIA controller when a GPU is later hot-plugged or returned from VFIO.

Fix: Add an `allow-no-gpus` initialization mode so NVML can initialize without a visible NVIDIA GPU and handle it when it becomes available.

### DIFF
--- a/lact-daemon/src/lib.rs
+++ b/lact-daemon/src/lib.rs
@@ -13,8 +13,9 @@ mod tests;
 use anyhow::Context;
 use config::Config;
 use futures::future::select_all;
+pub use lact_schema::NvidiaInitMode;
 use server::{Server, handle_stream, handler::Handler};
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 use std::{os::unix::net::UnixStream as StdUnixStream, time::Duration};
 use tokio::net::UnixStream;
 use tokio::runtime::LocalOptions;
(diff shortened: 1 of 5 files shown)


ilya-zlobintsev:
I don't quite understand why this is needed - NVML is initialized on-demand, specifically only when it finds a drm device with the `nvidia` driver. Is this about some reinit case when NVML is initialized and then all the nvidia GPUs are detached?

hurryman2212:
> I don't quite understand why this is needed - NVML is initialized on-demand, specifically only when it finds a drm device with the `nvidia` driver. Is this about some reinit case when NVML is initialized and then all the nvidia GPUs are detached?

I was in hurry too much for PR. The problem basically happened with some misunderstanding of `lactd` occupying the usage for NVIDIA DRM, preventing device remove or driver remove for VFIO passthrough. During the alteration of `lactd` code, I added `NO_ATTACH` and this regression was made. After that, I mispointed the root cause and sent this PR. This does not happen in the upstream. Sorry.

EDIT:

For anybody using multi NVIDIA GPUs, just stop `lactd`, register the device via VFIO, and restart `lactd` ...

PR #1119, opened 2026-07-18
feat: add `detach` & `attach` commands for multi-GPU environments without losing current settings briefly

hurryman2212:
Problem: Releasing one GPU for driver unbind or VFIO requires stopping `lactd`, which resets the settings of all managed GPUs to their defaults. (the current `lactd`'s behavior)

Fix: Add per-GPU `detach` and `attach` commands that release only the selected GPU and reapply its saved configuration when it re-attaches.

Example usage:

```
lact cli --gpu-id 1 detach # Or: lact cli --gpu-id 10DE:2B85-10DE:2057-0000:03:00.0 detach
# Pass it to VFIO driver, and VM via virsh
# ... (playing Windows games)
# Reclaim it from VM via virsh, then rebind it to NVIDIA
lact cli --gpu-id 1 attach # Or: lact cli --gpu-id 10DE:2B85-10DE:2057-0000:03:00.0 attach
# No other GPU will have intermediate and brief reset in their LACT settings.
```

(otherwise, just make `lactd` not to reset settings for all GPUs when the service is just stopped (normally), which can be more desirable and undesirable by the perspectives)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Linux GPU Control Application
+
 <a href="https://translate.fedoraproject.org/engage/lact/">
 <img src="https://translate.fedoraproject.org/widget/lact/svg-badge.svg" alt="Translation status" />
 </a>
@@ -9,7 +10,7 @@ This application allows you to control your AMD, Nvidia or Intel GPU on a Linux
 system.
 
 | GPU info                          | Overclocking                      | Fan control                       |
-| ----------------------------------| ----------------------------------| ----------------------------------|
+| --------------------------------- | --------------------------------- | --------------------------------- |
 | ![image](./res/screenshots/1.png) | ![image](./res/screenshots/2.png) | ![image](./res/screenshots/3.png) |
 | Software info                     | Historical data                   |                                   |
 | ![image](./res/screenshots/4.png) | ![image](./res/screenshots/5.png) |                                   |
@@ -49,7 +50,7 @@ The service can also be used standalone with a config file, for example in headl
 - [Installation](#installation)
 - [Hardware support](https://github.com/ilya-zlobintsev/LACT/wiki/Hardware-Support)
 - [Frequently asked questions](https://github.com/ilya-zlobintsev/LACT/wiki/Frequently-asked-questions)
-- [Enable overclocking on AMD](https://github.com/ilya-zlobintsev/LACT/wiki/Overclocking-(AMD))
+- [Enable overclocking on AMD](<https://github.com/ilya-zlobintsev/LACT/wiki/Overclocking-(AMD)>)
 - [Config file reference](./docs/CONFIG.md)
 - [API](./docs/API.md)
 - [Power profiles daemon note](#power-profiles-daemon-note)
@@ -68,6 +69,7 @@ The service can also be used standalone with a config file, for example in headl
 
   It is only available on Debian 12+ and Ubuntu 22.04+ as older versions don't
   ship gtk4.
+
 - Fedora: use the
   [Copr repository](https://copr.fedorainfracloud.org/coprs/ilyaz/LACT/), or
   download an RPM from
@@ -80,12 +82,14 @@ The service can also be used standalone with a config file, for example in headl
 
   Only tumbleweed is supported as leap does not have the required dependencies
   in the repos.
+
 - NixOS: There is a package available in
   [nixpkgs](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=lact).
 - Solus: Available in the offical repository: `eopkg it lact`
 - Flatpak (universal): Available on [Flathub](https://flathub.org/apps/io.github.ilya_zlobintsev.LACT) and in [releases](https://github.com/ilya-zlobintsev/LACT/releases/).
 
   See the [Flatpak documentation](./flatpak/README.md) for additional notes.
+
 - Docker (service only, no GUI): See [DOCKER.md](./docs/DOCKER.md)
 - Build from source.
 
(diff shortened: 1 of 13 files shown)


ilya-zlobintsev (changes requested):
This is a huge amount of logic for an obscure feature... a better solution for your problem would be figuring out why LACT keeps the GPU from being detached, as it should not do that (my suspicion is the nvapi file handle being open) 

But, even if that is not possible, there are much simpler ways to implement this. And if you are not sure about how to do it, it's better to just send a feature request, rather than submitting an overly complex change that your LLM generated for you.

hurryman2212:
> This is a huge amount of logic for an obscure feature... a better solution for your problem would be figuring out why LACT keeps the GPU from being detached, as it should not do that (my suspicion is the nvapi file handle being open)
> 
> But, even if that is not possible, there are much simpler ways to implement this. And if you are not sure about how to do it, it's better to just send a feature request, rather than submitting an overly complex change that your LLM generated for you.

First of all, I want to clarify `attach` and `detach` entrance logic is specifically designed by me except Codex GPT 5.6 (xhigh) is used for the actual Rust code, as I don't use this PL. The code is fully tested by me without LLM involved. If the repository's policy (I couldn't find it) is against it, I will withdraw it.

Second, `lactd` on my environments keeps opening fd ownership on each DRM device (in my environments, `labwc` does not do EGL bindings on NVIDIA dGPUs other than my Ryzen iGPU), so without stopping it (as I already mentioned), it does not let each card to be handoff to VFIO. So that's the reason of "why LACT keeps the GPU from being detached".

Third, I chose this approach because this looks like most simplistic implementation to workaround the issue (other than stop service to drop settings such as power limit for brief time - although if you choose to drop reset stage for normal shutdown, that's fine, but like I said, that is optimal or sub-optimal choice depending on perspectives), than diagnosing whole code base for all locksteps requiring fd opened.

ilya-zlobintsev:
> although if you choose to drop reset stage for normal shutdown, that's fine, but like I said, that is optimal or sub-optimal choice depending on perspectives),

This could be implemented as a config option. There is already a `disable_clocks_cleanup` flag that was added a long time ago to work around some AMD issues, and there could easily be another option which disables cleanup for all settings.

> Second, lactd on my environments keeps opening fd ownership on each DRM device (in my environments, labwc does not do EGL bindings on NVIDIA dGPUs other than my Ryzen iGPU)

LACT keeps the dri files `(/dev/dri/render*)` open for AMD and Intel GPUs. On Nvidia those files are not used at all, but `/dev/nvidia*` ones are opened instead. Part of this is custom logic (NvAPI, as well as opening the file directly for raw kernel driver calls), and part is the NVML handle that cannot be avoided. However, in principle, LACT should not need to keep the handles open if it is not actively being interacted with (e.g. using custom fan control). It would be a nice thing to have, though likely not straightforward to cleanly implement.

The reason for my harsh response earlier is that the logic is overly complicated and touches much more than it should. Properly reviewing it to get to a mergeable form would take me more effort and time than it took you to prompt the LLM to make this, or compared to just implementing it manually myself. Which is why I'd prefer a feature request than to be dumped an unmaintainable patch.

hurryman2212:
> > Second, lactd on my environments keeps opening fd ownership on each DRM device (in my environments, labwc does not do EGL bindings on NVIDIA dGPUs other than my Ryzen iGPU)
> 
> LACT keeps the dri files `(/dev/dri/render*)` open for AMD and Intel GPUs. On Nvidia those files are not used at all, but `/dev/nvidia*` ones are opened instead.

Yeah, I meant that. My mistake.

> This could be implemented as a config option. There is already a `disable_clocks_cleanup` flag that was added a long time ago to work around some AMD issues, and there could easily be another option which disables cleanup for all settings.

> The reason for my harsh response earlier is that the logic is overly complicated and touches much more than it should. Properly reviewing it to get to a mergeable form would take me more effort and time than it took you to prompt the LLM to make this, or compared to just implementing it manually myself. Which is why I'd prefer a feature request than to be dumped an unmaintainable patch.

OK. But I still think the `disable_clocks_cleanup` is not optimal option as the ideal normal shutdown behavior is the opposite, and it seems that dynamically accept config change will generate "config change -> service stop -> detach -> config change -> service start" routines and it would look weird, if the multi-GPU users like me still want the service normal behavior when stops completely.

ilya-zlobintsev:
@hurryman2212 please try out #1123. It should work the same as this PR, without the code complexity.

hurryman2212:
> @hurryman2212 please try out #1123. It should work the same as this PR, without the code complexity.

With my RTX 5090 with +275(Core)/+6000(VRAM)MHz OC setting, `detach` and `reattach` command works (literally); `lactd` no longer sees a device within its routines when it detaches, and will see it when it comes back.

However, the mentioned issue still remains; `lactd` will keep occupying these even when the device is detached. The belows are the paths for which `lactd` does not let the handles go:  
  - /dev/nvidia*
  - /dev/nvidiactl
  - /dev/nvidia-caps/nvidia-cap1
  - /dev/nvidia-caps/nvidia-cap2

ilya-zlobintsev:
@hurryman2212 I've updated the PR, please try it again. On my system, detaching the nvidia card drops all the nvidia FDs now.
Note that this might not be the case if you have multiple nvidia cards present, and are only detaching one, as NVML might keep a handle on the detached card by itself even when it is unused.

hurryman2212:
> @hurryman2212 I've updated the PR, please try it again. On my system, detaching the nvidia card drops all the nvidia FDs now. Note that this might not be the case if you have multiple nvidia cards present, and are only detaching one, as NVML might keep a handle on the detached card by itself even when it is unused.

Now the problem is resolved. Ty!

PR #1132, opened 2026-08-01
refactor: gpu stats section

makarov-roman:
I decided to move these changes from power_page PR, so I can proceed with stats toggles. 

I think we can skip this PR for this release


ilya-zlobintsev (changes requested):
Looks generally fine, and duplicating this is probably simpler than the previous attempts at unifying GpuStat with the graphs.

But, there is a small behaviour regression: the thermals page used to show "no fan detected", it doesn't anymore, I think it's a useful thing to keep

Image

Image

ilya-zlobintsev (approved):
LGTM, but please avoid force pushing after review - this practically means all of the changes have to be re-reviewed

PR #1135, opened 2026-08-02
feat: replace nvml with nvapi for nvidia throttle info

makarov-roman:


makarov-roman:
I tried to OC my gpu from this branch, and tbh, the change is not very useful. My pedantic autistic part likes what I see vRel, but I couldn't find real use case to expose that. Basically it shows vRel every time once you have have load and not throttled by power/temps. 

If you agree, I'll close it

PR #1145, opened 2026-08-08
fix: use consistent reset/default terminlogoy

makarov-roman:


lact-gui/i18n/en/lact_gui.ftl
```diff
@@ -56,7 +56,8 @@ min-fan-speed = Minimum Fan Speed (%)
 zero-rpm = Zero RPM
 zero-rpm-stop-temp = Zero RPM stop temperature (°C)
 static-speed = Static Speed (%)
-reset-button = Reset
+reset-button = Reset Now
```
ilya-zlobintsev: I'm not sure about this. The "now" label makes sense in the OC section, but this string is also used in the graphs config and in the reset dialog. I think there it's redundant to say, as IMO it's pretty clear in their context that it's an instant operation

makarov-roman: There are 4 cases and all of them fit imo. Red button means it's destructive, nothing signals that the action is applied immediately.

makarov-roman: I went through UI again - and agree.
1. grpahs - use default with destructive class
2. OC and Thermals still reset now
3. reset all config dialog - normal reset

ilya-zlobintsev (approved):

PR #1155, opened 2026-08-14
feat: report clockspeed of extra clock domains on nvidia

ilya-zlobintsev:
image

Currently, the only way to see these in the UI is through configuring graphs (same as AMD extra clocks)

### DIFF
--- a/lact-daemon/src/server/gpu_controller/nvidia.rs
+++ b/lact-daemon/src/server/gpu_controller/nvidia.rs
@@ -6,7 +6,9 @@ use crate::{
     bindings::nvidia::NvPhysicalGpuHandle,
     server::gpu_controller::{
         common::{fan_control::FanCurveExt, resolve_process_name},
-        nvidia::nvapi::{CLOCK_CLIENT_CLK_VF_POINT_TYPE_PROG, ClockClientClkVfPointInfoV1},
+        nvidia::nvapi::{
+            CLOCK_CLIENT_CLK_VF_POINT_TYPE_PROG, ClockClientClkVfPointInfoV1, NvGpuClockDomainId,
+        },
     },
 };
 use amdgpu_sysfs::{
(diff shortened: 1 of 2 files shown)


lact-daemon/src/server/gpu_controller/nvidia.rs
```diff
@@ -1048,13 +1067,7 @@ impl GpuController for NvidiaGpuController {
...
                 vram_clockspeed: device.clock_info(Clock::Memory).map(Into::into).ok(),
                 target_gpu_clockspeed: None,
-                sensors: [
-                    ("SM", device.clock_info(Clock::SM)),
-                    ("Video", device.clock_info(Clock::Video)),
-                ]
-                .into_iter()
-                .filter_map(|(label, result)| Some((label.to_owned(), result.ok()?.into())))
-                .collect(),
+                sensors: extra_clocks,
```
makarov-roman, line 1070: I think a significant part of complexity we have around gpu stats is this pattern, where some stats are defined in schema, and some are dynamic. Imo it should be one or another.

ilya-zlobintsev: The problem is that we have two kinds of stats (clockspeeds and temps):
- A few well-known mostly-universal ones (GPU/VRAM) clock, that are cross-vendor and the UI is expected to show them front and center
- A bunch of additional, vendor-specific ones, which are less important in most cases

For the first one, we care about them a lot, and the UI needs to know about them, so the best way for that is a static schema. 
For the rest, the distinction is mostly irrelevant for UI purposes - we don't need to bother with displaying XBAR vs HUB vs FCLK clocks differently. So even the schema would be static (with vendor-specific extensions), the UI code would likely still handle it just as label-value pairs.

makarov-roman: hm, It might be fine, but what's downside describing them in the schema? It makes UI more predictable. e.g. remember bug when intel started to expose much more temps and it broke stats layout.

ilya-zlobintsev: I don't think describing them in a schema provides much value, since the client will still dynamically put them in a list/map - might as well cut out the middle part. It would have the benefit of defining what potentially might be there, but I don't think that is too useful atm

PR #1157, opened 2026-08-14
feat: fmt gpu name in gpu-picker

makarov-roman:
image

### DIFF
--- a/lact-gui/src/app/utils/formatting.rs
+++ b/lact-gui/src/app/utils/formatting.rs
@@ -123,6 +123,31 @@ pub fn fmt_timestamp_to_dt(timestamp_ms: &i64) -> String {
     date_time.strftime("%H:%M:%S").to_string()
 }
 
+const GPU_MODEL_SERIES: &[&str] = &[
+    "RTX", "GTX", "GTS", "GT", "RX", "R9", "R7", "R5", "HD", "Arc",
+];
+
+pub fn fmt_clean_gpu_name(name: &str) -> &str {
+    let mut short = name.trim();
+
+    // PCI device names carry the marketing name in brackets, e.g. "DG2 [Arc A770]"
+    if let Some((_, bracketed)) = short.split_once('[')
+        && let Some((model, _)) = bracketed.split_once(']')
+    {
+        short = model.trim();
+    }
+
+    let mut model = short;
+    while let Some((word, rest)) = model.split_once(' ') {
+        if GPU_MODEL_SERIES.contains(&word) {
+            return model;
+        }
+        model = rest;
+    }
+
+    short
+}
+
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 #[repr(u8)]
 pub enum ByteUnit {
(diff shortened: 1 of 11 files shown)


lact-gui/src/app/utils/formatting.rs
```diff
@@ -312,12 +331,40 @@ mod tests {
...
+    #[test]
+    fn fmt_clean_gpu_name_keeps_names_without_a_series() {
         assert_eq!(
-            fmt_timestamp_to_dt_with_tz(&timestamp_ms, jiff::tz::TimeZone::UTC),
-            "00:00:00"
+            fmt_clean_gpu_name("AMD Radeon 780M Graphics"),
+            "AMD Radeon 780M Graphics"
         );
+        assert_eq!(fmt_clean_gpu_name("Phoenix1"), "Phoenix1");
+    }
```
ilya-zlobintsev: We could probably strip the vendor name even in cases where the rest of the pattern matching doesn't work?

lact-gui/src/app/utils/formatting.rs
```diff
@@ -119,14 +119,33 @@ pub fn fmt_clockspeed(clock_mhz: Option<u64>, ratio: f64) -> String {
...
+    let mut model = short;
+    while let Some((word, rest)) = model.split_once(' ') {
+        if GPU_MODEL_SERIES.contains(&word) {
+            return model;
+        }
+        model = rest;
+    }
+
+    short
 }
```
ilya-zlobintsev: I think there is benefit to moving it daemon-side, so that these names are used in the device listing universally (e.g. in CLI tool)

lact-gui/src/app/utils/formatting.rs
makarov-roman: agree.
Overall https://www.techpowerup.com/gpu-specs uses good approach.
AMD/Nvidia stripped. GeForce/Radeon kept.
Initially I wanted to have something like RTX 4070 Super, which looks good, but there are some cases where it doesn't. E.g. Radeon VII would be just VII

lact-gui/src/app/utils/formatting.rs
makarov-roman: I think it better fits schema, because daemon shouldn't own formatting. So the clients will have access to raw data if needed

lact-schema/src/lib.rs
```diff
@@ -183,13 +210,14 @@ impl DeviceInfo {
...
-            .unwrap_or("Unknown")
-            .to_owned();
+        let mut gpu_model = clean_gpu_name(
+            self.drm_info
+                .as_ref()
+                .and_then(|drm| drm.device_name.as_deref())
+                .or_else(|| pci_info.and_then(|pci_info| pci_info.device_pci_info.model.as_deref()))
+                .unwrap_or("Unknown"),
+        )
+        .to_owned();
```
ilya-zlobintsev: I think the cleanup should not be done for `info_elements`. With this, we never show the full name anywhere. I think the info page warrants the full name, especially since the short one can be quite limited at times:

Image

ilya-zlobintsev (changes requested):

lact-schema/src/lib.rs
makarov-roman: just to make sure I understand you - the clean name should be used in `lact cli list` but not `lact cli info`?  and on gui it would be gpu picker with clean name and HWInfo -> GPU model with full

ilya-zlobintsev: Correct

ilya-zlobintsev (approved):

Issue #1147, opened 2026-08-10
Feature: Runtime XBAR clock and per-domain MSVDD control on NVIDIA Blackwell


Loong0x00:
## Summary

I found working runtime controls for the XBAR clock offset and the XBAR-domain MSVDD voltage offset on NVIDIA Blackwell under Linux. Both controls can be read, written, read back exactly, measured through the hardware clock query, and restored without a reboot.

This exposes an additional clock domain that is currently not adjustable in LACT or the public NVIDIA Linux tools. It can provide measurable performance gains in workloads sensitive to the GPU crossbar/cache/memory path. The voltage control is independent from the frequency control and becomes relevant at larger XBAR offsets.

The controls were tested through the private RM ClockClient interface on an RTX 5090 with driver 610.57.04. A minimal standalone C demonstration is available here:

https://gist.github.com/Loong0x00/959d7e934366a721399a84e7943cf442

## Controls

The following R610 controls were used:

| Command | ID | Parameter size |
|---|---:|---:|
| `NV2080_CTRL_CMD_CLK_CLK_DOMAINS_GET_INFO` | `0x20809019` | `0x3030` |
| `NV2080_CTRL_CMD_CLK_CLK_DOMAINS_GET_CONTROL` | `0x2080901b` | `0x083c` |
| `NV2080_CTRL_CMD_CLK_CLK_DOMAINS_SET_CONTROL` | `0x2080d01c` | `0x083c` |
| `NV2080_CTRL_CMD_CLK_MEASURE_FREQ` | `0x20809006` | `0x0008` |

The relevant layout of the `0x83c` control block on this driver is:

```text
+0x04  controllable-domain mask (0x000000ff)

domain header: 0x3c bytes
domain stride: 0x40 bytes
XBAR domain entry: index 1, base +0x7c

+0x84  frequency-offset mode
+0x88  signed frequency offset in kHz
+0x90  signed XBAR-domain MSVDD offset in microvolts
```

`CLK_MEASURE_FREQ` domain `2` returns the physical XBAR clock in kHz. The reported XBAR frequency-offset range on this system is `-1000` to `+1000 MHz`.

This layout is private and driver-branch-specific. The result above is for R610; it should not be assumed to match R570 or future branches without checking the returned structures.

## Minimal demonstration

Build:

```bash
cc -std=gnu11 -O2 -Wall -Wextra -Wpedantic -Werror \
  xbar_clock_demo.c -o xbar_clock_demo
```

[...]

PR #1158, opened 2026-08-14
feat: adjustable XBAR clock and per-domain MSVDD offsets on nvidia (Turing and newer)

Panchovix:
## Summary

Adds runtime control of the clock domains that NVML does not expose on Nvidia,
implementing #1147. Each adjustable domain gets a frequency offset and an offset
on its own MSVDD rail. On Blackwell this makes XBAR tunable, which has a
measurable effect on workloads that lean on the crossbar/cache/memory path.

Rebased onto master now that #1155 has merged, so this is a single commit.

## How

The offsets go through the RM clock domain board object group
(`NV2080_CTRL_CMD_CLK_CLK_DOMAINS_GET_INFO` / `GET_CONTROL` / `SET_CONTROL`),
reusing the existing `DriverHandle` RM ioctl client rather than adding a second
low-level interface.

NvAPI is not usable here even though LACT already links it. On R610 the domains
report board object type `0x10`, which its V1 control struct skips in both
directions — it copies neither the offsets it reads nor the ones you ask it to
write. Its `GetInfo` returns entries that are entirely zero, so the domain
identity needed to locate XBAR is not recoverable through it at all. For the
record, the NvAPI entry points are `0xF58938F5` (ClkDomainsGetControl) and
`0xD14B69CF` (ClkDomainsSetControl), found by matching the library's internal
QueryInterface table against those RM command constants.

The XBAR index is not hardcoded the way the issue's demo does it. `GET_INFO` is
read first: each entry carries `1 << domain id` plus the allowed offset range,
so the control index is discovered and the range is enforced before anything is
written. Writes are read-modify-write, since the RM control replaces the whole
group in one transaction.

The layout of an info entry moves with its board object type, which is the first
byte of the entry. Two are known: Blackwell reports `0x10`, and the generations
before it report `0x0b` and put everything after the parent object four bytes
further back. The offset range is read positionally behind an explicit check for
that type, so an entry of an unknown future type is left alone and fails safe
instead of being read at a guessed offset.

Domains NVML already drives (GPC, memory) are excluded too, so these offsets
cannot conflict with the pstate offsets LACT applies through NVML — see the
interaction described in #936. On an RTX 5090 that leaves XBAR, SYS and Power.

## Testing

Verified on an RTX 5090, driver 610.57.04, VBIOS 98.02.2E.80.C9, under load at
100% utilisation and 475 W. The figures are the RM-measured physical XBAR clock,
not a reported target:

| XBAR request | measured XBAR (mean) | delta |
|---|---|---|
| 0 | 2489 MHz | baseline |
| +200 MHz | 2680 MHz | +191 MHz |
| +400 MHz | 2755 MHz | +266 MHz (power limited) |
| 0 | 2480 MHz | back to baseline |

Frequency and MSVDD offsets both read back exactly and reset cleanly. Worth
knowing for anyone testing this: the offset does nothing at idle or under
`nvidia-smi` locked clocks, because it shifts the domain's V/F curve and the
pstate pins the clock until there is real load.

`driver.rs` carries an `#[ignore]`d round-trip test; run it with
`LACT_TEST_GPU_MINOR=<n> cargo test -- --ignored`. It skips itself when the
variable is unset.

## Caveats

- The parameter layout is private and branch-specific. It was derived on R610
  and should not be assumed to hold on other branches without re-checking the
  returned structures. Every field is validated against `GET_INFO` first and the
  block sizes are asserted at compile time, so a mismatch should surface as a
  failed control rather than a bad write.
- A successful write with exact readback does not imply stability. Higher XBAR
  offsets need voltage to stay clean, and the useful setting is workload- and
  power-budget-dependent.
- Nothing in the interface reports a range for the per-domain rail offsets, and
  reading the control block back only echoes whatever was written, so there is no
  limit to derive and no way to observe whether a board honoured an offset. MSVDD
  is therefore capped at ±250 mV purely as a guard rail, matching the range the
  GUI already uses for the AMD voltage offset. Happy to change that number.
- MSVDD reads back correctly but its physical effect could not be confirmed
  here, as the rail is not readable. On workstation parts it appears to be
  clamped by VBIOS.
- Tested on GB202 (RTX 5090, RTX PRO 6000 Blackwell) and GB203 (RTX PRO 4000
  Blackwell SFF, RTX 5070 Ti).
- The older `0x0b` entry layout was measured on an RTX 2000 Ada, which does
  respond to these offsets: with the core clock pinned at 2430MHz by a memory
  copy, +300MHz on XBAR took it from 2130 to 2430MHz and -300MHz took it to
  1830MHz, returning to 2130MHz when cleared. SYS followed XBAR both ways. An
  earlier revision of this PR excluded pre-Blackwell on the grounds that the
  effect was unproven; that is no longer the case.

## Clock domain reference

The domains below come from `NV2080_CTRL_CMD_CLK_CLK_DOMAINS_GET_INFO`. Frequencies
were sampled with `NV2080_CTRL_CMD_CLK_MEASURE_FREQ` (mask `1 << domain id`), which
returns the *physical* clock rather than a reported target.

| Domain | Adjustable | Offset range | RTX 5090 (idle) | RTX PRO 6000 (under load) |
|---|---|---:|---:|---:|
| GPC | yes | ±1000 MHz | 147 MHz | 2364 MHz |
| XBAR | yes | ±1000 MHz | 1056 MHz | 2262 MHz |
| SYS | yes | ±1000 MHz | 915 MHz | 1703 MHz |
| Memory | yes | −1000/+3000 MHz | 404 MHz | 16000 MHz |
| Power | yes | ±1000 MHz | 832 MHz | 1563 MHz |
| HUB | no | — | 541 MHz | 540 MHz |
| Display | no | — | 675 MHz | 675 MHz |
| Legacy | no | — | 1859 MHz | 1705 MHz |
| MSD | no | — | 0 | 0 |

- **GPC** — the shader/graphics partition clock, i.e. what NVML and every tool
  reports as "GPU clock".
- **XBAR** — the crossbar between the SMs and the L2/memory partitions. This is
  the domain issue #1147 is about.
- **SYS** — the system partition: host interface, front-end and the parts of the
  uncore outside the shader cores. This is the other domain measured in the
  NVIDIA report linked from #1147 (NVIDIA/open-gpu-kernel-modules#1266), because
  a finite memory-clock maximum can constrain XBAR and SYS together.
- **Memory** — the DRAM clock, already driven through NVML.
- **Power** — named after the NvAPI clock domain enum (`Pwr = 20`); the exact
  block it clocks is not documented here. Conventionally PWRCLK is the power
  management microcontroller's clock, but the measured behaviour does not quite
  fit that: it scales with load and holds a near-constant ratio to SYS (1.10 on
  the 5090 at idle, 1.09 on the PRO 6000 under load), which looks more like an
  uncore domain coupled to SYS. Treat the label as unverified.
- **HUB / Display / Legacy / MSD** — report a zero offset range, so they are
  observable but not adjustable. They are deliberately not surfaced in the UI.

GPC and Memory are adjustable at the RM level but excluded from this PR, since
LACT already drives both through NVML pstate offsets and the two paths conflict
(see #936).

## On the two voltage rails

The per-domain control block carries four voltage rail offsets. Rail 0 is NVVDD
and rail 1 is MSVDD; this PR only exposes rail 1.

**NVVDD** is the main core rail, and it is already visible in LACT — it is the
value shown as GPU voltage, the X axis of the V/F curve editor, and what the
voltage boost control acts on.

**MSVDD** powers the memory-system/uncore logic, separate from both NVVDD and the
DRAM's own supply. It appears to have no read path: `ClientVoltRailsGetStatus`
(`0x465F9BCF`) only accepts a struct with a single rail entry (`rail_id = 0`), and
requesting 2, 3 or 4 rails returns `NVAPI_INCOMPATIBLE_STRUCT_VERSION`. That is why
the UI can set an MSVDD offset but cannot display the resulting voltage, and why
the offset's physical effect is reported here as unconfirmed rather than verified.

## Pictures

RTX 5090 Pictures
XWin_MobaX_iJAR4diDSP

XWin_MobaX_vUTNdyjKub

## Performance difference

On a RTX 6000 PRO MaxQ, applying only +550Mhz on XBAR clocks, on a diffusion task (txt2img, on forge neo, Anima, 896x1088, batch size 4, 25 steps), it decreases gen time by about 10% when not power limited, and about 2-3% when power limited (so there's still a difference when power limited).

Max Q Picture
imagen

## Unrelated observation

`tests::snapshot_everything` fails on any machine that has a real GPU at
`0000:09:00.0`: the mocked RTX 4080 fixture declares that PCI slot and NVML
resolves it to the real card, so the test reads real hardware. It fails
identically on an unmodified checkout, so it is pre-existing and not caused by
this change, but it does mean the nvidia snapshot must not be regenerated on
such a machine.

Thanks to @Loong0x00 for finding the control and characterising it in #1147.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


SHANAjam:
Hi,

I tested this on Windows (RTX 5090, driver 610.62). The V1 ClkDomains struct
is indeed unusable, but the V2 struct (version 0x000261A4) works.

I confirmed:
- ClkDomainsGetControl 0xF58938F5
- ClkDomainsSetControl 0xD14B69CF
- XBAR domain index 1, offset +0x114, MSVDD +0x11C
- short SET/readback/restore passed

Note: I am an AI assistant helping the repository owner. This is AI-assisted
and only tested on one configuration.

Details: https://github.com/SHANAjam/rtx5090-xbar-control
(docs/TECHNICAL_NOTES.md)

Panchovix:
Update at how it look right now. Let me know if it's too cluttered to add another section maybe, with these per-domain offsets.

imagen

Panchovix:
Update as how it looks right now, these options are moved to advanced, so it doesn't clutter the main settings.

imagen

imagen

The issue I have right now is aligning the text correctly.

Panchovix:
Update as how it looks right now, these options are moved to advanced, so it doesn't clutter the main settings.

imagen

XWin_MobaX_AInCKvqsKS

Loong0x00:
I explored a different, experimental model for the same Blackwell controls and
published it here for reference:

- Branch: https://github.com/Loong0x00/LACT/tree/blackwell-multidomain-vf
- English overview, verification boundary, and AI disclosure: https://github.com/Loong0x00/LACT/blob/blackwell-multidomain-vf/docs/BLACKWELL_MULTI_DOMAIN_VF_EXPERIMENT.md
- Detailed experiment/development ledger (Chinese): https://github.com/Loong0x00/LACT/blob/blackwell-multidomain-vf/docs/BLACKWELL_MULTI_DOMAIN_VF_DEVELOPMENT_GUIDE.md

The alternative data/UI model is one MSVDD view containing independently
visible XBAR, SYS, and NVD 127-point STATUS curves. It keeps domain-wide
frequency offsets, XBAR/SYS point CONTROL, the three native per-domain MSVDD
mapping offsets, and the GPC-to-XBAR propagation ratio as separate layers. A
linked MSVDD control is only a convenience action over the three native mapping
offsets. Literal RM `targetFreq` / `actualFreq` fields and the independent
`CLK_MEASURE_FREQ` counter are displayed separately rather than being renamed
to a final grant or physical frequency.

Writes are bound to the exact driver/VBIOS ROM/NVIDIA API/GSP/topology contract,
use full-object preimages and exact readback, and are not automatically applied
after any contract change. NVD point writes remain disabled because fresh
CONTROL/STATUS discarded both tested directions; propagation SET/readback is
also not claimed as proof that the relation is the binding constraint for every
workload.

This implementation and most of its documentation were generated by OpenAI
Codex under my direction, then built and tested on my RTX 5090 with driver
610.57.04 and VBIOS 98.02.2E.80.50. The tested MSVDD `-50..+50 mV` range is only
an adoption/restore envelope, not a vendor safe range or stability claim.

I am posting this as an alternative design/research reference, not as a request
to replace or merge it into this PR.

ilya-zlobintsev:
I'd like to test some approaches with NvApi myself before this.

PR #1161, opened 2026-08-16
fix: swtich nvidia V/F curve to offsets

makarov-roman:
I tried to push my gpu further with VF curve and encountered annoying problem. When you apply curve while gpu under the load - some voltage points jump +-1 clocks step, and when you apply the curve multiple times the error becomes cumulative. E.g. after 3-4 applies (with shift), you can see deviation of 45-60mhz from what you tried to set. 

Even applying default results in some points going +- 15mHz. I'll explain in comments what I found


lact-daemon/src/server/gpu_controller/nvidia.rs
```diff
@@ -394,138 +395,54 @@ impl NvidiaGpuController {
...
-                let base_freq: i32 = if let Some(base_points) = &*base_vf_curve {
-                    let base_point = base_points
-                        .iter()
-                        .find(|point| point.index as usize == i)
-                        .expect("Mismatched base point");
-                    base_point.base_freq.cast_signed() * 1000
-                } else {
-                    current_point.vf_tuple_base.freq_khz.try_into()?
-                };
-                let offset_khz = configured_mhz * 1000 - base_freq;
```
makarov-roman, line 514: this is the problem, base_freq value is not static and changes under load, which leads to error in offset calculation. 

I couldn't find a way to get reliable static value. (and I really tried just to avoid schema change).

lact-daemon/src/config.rs
```diff
@@ -237,6 +237,18 @@ impl Config {
                         }
                     }
                 }
+                7 => {
```
makarov-roman, line 244: I decided to skip migration. technically we can support it, but it sounds like unreliable migration of unreliable curve. It's easy to miss some defect when you have hundreds of points.

lact-daemon/src/server/gpu_controller/nvidia.rs
```diff
@@ -627,6 +544,102 @@ fn vf_curve_point_is_editable(point: ClockClientClkVfPointInfoV1) -> bool {
...
+            let current_mv = status.vf_points[i].voltage_uv / 1000;
+            if configured_mv != current_mv {
+                warn!(
+                    "voltage is immutable - point {i} is at {current_mv}mV but was configured as {configured_mv}mV, skipping it"
+                );
+                continue;
+            }
+        }
+
+        let offset_mhz = configured_point.clockspeed_offset;
```
makarov-roman, line 623: offset is much more reliable control. Even though base_freq can be not in sync between UI/daemon/driver - we'll always get the correct offset. UI render of the curve might be a bit off, e.g. when voltage point says 2970, but actual clock on the same point is 2955

makarov-roman: maybe we can add some polling when curve is opened to improve accuracy

lact-schema/src/lib.rs
```diff
@@ -524,6 +524,7 @@ pub struct NvidiaVfPoint {
     pub voltage: u32,
     pub base_freq: u32,
     pub base_voltage: u32,
+    pub freq_offset: i32,
```
ilya-zlobintsev, line 528: This should be Option or have `#[default]`. Otherwise the UI will fail to deserialize the response when running with an old daemon.
We don't need to maintain proper compatibility between mismatched client and daemon, but this will likely end up with an ugly error

lact-schema/src/config.rs
```diff
@@ -146,6 +152,13 @@ pub struct CurvePoint {
     pub clockspeed: Option<i32>,
 }
 
+#[skip_serializing_none]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, Default, PartialEq, Eq)]
+pub struct NvidiaCurvePoint {
+    pub clockspeed_offset: i32,
+    pub voltage: Option<u32>,
```
ilya-zlobintsev, line 159: Do we need the voltage in the config, when it is only used for validation? I understand the UI can make use of this as a sanity check, but if you consider this just as a config - this option does not make much sense, as you can never edit it

lact-daemon/src/server/gpu_controller/nvidia.rs
```diff
@@ -629,6 +544,102 @@ fn vf_curve_point_is_editable(point: ClockClientClkVfPointInfoV1) -> bool {
...
+/// The base curve for a point. Changes with GPU state(load,temps, e.tc), so it may only be used for display and validation
+fn vf_curve_point_base_freq_khz(
+    status: &ClockClientClkVfPointsStatusV3,
+    control: &ClockClientClkVfPointsControlV1,
+    i: usize,
+) -> i32 {
+    if status.b_vf_tuple_base_supported == 0 {
+        // Cards without the base tuple (Turing) only expose the resulting curve
+        status.vf_points[i].freq_khz.cast_signed()
+            - unsafe { control.vf_points[i].data.prog.freq_offset_khz }
```
ilya-zlobintsev, line 556: I don't remember this 100%, but IIRC the offset value reported here is pre-"snapping" (offsets typically get snapped to the nearest 5mhz multiple), while the resulting frequency is post-snap. So it's probably safer to keep the cached based curve, even if this is only used for display

lact-daemon/src/server/gpu_controller/nvidia.rs
```diff
@@ -629,6 +544,102 @@ fn vf_curve_point_is_editable(point: ClockClientClkVfPointInfoV1) -> bool {
...
+        // Only report configurable and voltage-based points
+        if !vf_curve_point_is_editable(info.vf_points[i]) {
+            continue;
+        }
+
+        let point = status.vf_points[i];
+        let offset_khz = unsafe { control.vf_points[i].data.prog.freq_offset_khz };
+
+        let base_freq_khz = vf_curve_point_base_freq_khz(status, control, i);
+        let base_voltage_uv = if status.b_vf_tuple_base_supported == 0 {
```
ilya-zlobintsev, line 580: Same concern here regarding not storing the base curve

ilya-zlobintsev (changes requested):

lact-schema/src/config.rs
makarov-roman: I'm a bit confused by that. It's indeed used in [the validation](https://github.com/ilya-zlobintsev/LACT/pull/1161/changes#diff-dc5ac963ef64fa66ecf5c4ea4ce6520f82c6943769cc65701599ce3a75c24b38R615). I don't see how to replace it easily. imo the validation itself makes sense if somebody edited the config manually.

lact-daemon/src/server/gpu_controller/nvidia.rs
```diff
@@ -629,6 +544,99 @@ fn vf_curve_point_is_editable(point: ClockClientClkVfPointInfoV1) -> bool {
...
+            ensure!(
+                configured_mv == current_mv,
+                "Voltage is immutable - point {i} is at {current_mv}mV but was configured as {configured_mv}mV"
+            );
+        }
+
+        let offset_mhz = configured_point.clockspeed_offset;
+        let base_freq_mhz = vf_curve_point_base_freq_khz(status, &control, i) / 1000;
+
+        let point_min_offset_mhz = cmp::min(min_offset_mhz, -base_freq_mhz);
```
ilya-zlobintsev: Shouldn't this be `max`? E.g. we are editing a point with the base clock `500mhz`, allowed offset range is -1000 to 1000,  we should not allow more than -500 offset

ilya-zlobintsev (changes requested):

Panchovix:
Thanks for this! I was having the same issue and could cause issues after some reboots.

Issue #1159, opened 2026-08-16
Feature follow-up: expose remaining NVIDIA Blackwell clock topology and XBAR V/F status


Loong0x00:
## Scope

#1158 already implements generic frequency/MSVDD offsets for every non-NVML clock domain whose `CLK_CLK_DOMAINS_GET_INFO` entry reports a non-zero offset range. This follow-up deliberately does not duplicate that work. It tracks the other Blackwell clock controls and read-only state that have now been identified and tested on Linux.

Test platform: RTX 5090 / GB202, driver and GSP 610.57.04. All layouts below are private and branch-specific.

## Offset-capable domains currently reported

The current `0x20809019` response contains five entries with non-zero ranges:

| raw domain ID | mask | range | current handling |
|---:|---:|---:|---|
| 0 | `0x00000001` | -1000..+8000 MHz | GPC, already handled through NVML |
| 1 | `0x00000002` | -1000..+1000 MHz | XBAR, covered by #1158 |
| 4 | `0x00000010` | -1000..+5000 MHz | memory, already handled through NVML |
| 2 | `0x00000004` | -1000..+1000 MHz | SYS, covered by #1158 |
| 20 | `0x00100000` | -1000..+1000 MHz | covered generically by #1158; the human-readable name should remain marked unverified until the RM/NvAPI naming difference is resolved |

The remaining active entries report a zero offset range and should stay read-only.

## 1. Direct physical clock measurement

`CLK_MEASURE_FREQ` (`0x20809006`, 8 bytes) accepts a one-hot domain mask and returns the measured hardware frequency in kHz. This is not a requested clock or V/F target. It works for at least GPC, XBAR, SYS, MCLK and the other active domains on this card.

This would be useful as generic per-domain monitoring and as independent verification for every advanced control below.

## 2. Runtime-writable GPC to XBAR propagation ratio

The active topology contains a bidirectional type-3 GPC2CLK to XBARCLK relationship. Its U16.16 ratio is readable, writable, exactly readable after SET, measurable in hardware and restorable at runtime.

| operation | command | size |
|---|---:|---:|
| relationship info | `0x20809081` | `0x1518` |
| relationship control | `0x20809083` | `0x0c18` |
| internal relationship SET | `0x2080d084` | `0x0c18` |

On this system the discovered relation is type 3, source domain 0, destination domain 1, bidirectional, with factory ratio `0xe660 / 65536 = 0.89990234375`.

[...]

PR #1162, opened 2026-08-17
feat: adjustable GPC clock propagation ratio on nvidia (XBAR, SYS, Power (clocks), Legacy)

Panchovix:
## Summary

Makes the GPC→XBAR clock propagation ratio adjustable, implementing section 2 of
#1159. Blackwell propagates the core clock to XBAR through a ratio that is 0.9 from
the factory; raising it decouples XBAR from the core clock, and it is a separate
lever from the per-domain offset in #1158.

Draft because it is stacked on #1158, which adds the clock domain plumbing this
reuses. Until that merges this PR shows its commit too; happy to rebase once it
lands.

## How

The relationship lives in the RM clock propagation topology
(`CLK_TOP_PROP_RELS_GET_INFO` `0x20809081`, `GET_CONTROL` `0x20809083`,
`SET_CONTROL` `0x2080d084`), which is a board object group like the clock domains,
so it goes through the same `DriverHandle` RM ioctl client.

Following the procedure #1159 asks for: the relationship is discovered from the
info response rather than assumed, by looking for a ratio-carrying type whose
endpoints resolve to the GPC and XBAR domains. The entire control block is kept as
the restoration preimage, only that one `U16.16` field is replaced, and every other
byte is carried through untouched. A write that does not read back exactly is
rejected and the original block restored. In practice a change moves exactly three
bytes of the 3096 byte block.

Two things worth recording that are not in the issue:

- **Several ratio relationships can exist.** The Blackwell cards here report five
  GPC→XBAR ones, at indices 0, 10, 11, 12 and 13, with factory ratios of 0.9, 0.8,
  2.0, 1.2 and 1.5. What selects between them is not known, so only the first is
  ever touched and the rest are left alone.
- **The info response keeps reporting the factory ratio after a write.** That is
  the only way back to stock once the control block has changed, and it is what
  reset uses, so stock survives a daemon restart.

Pre-Blackwell cards describe GPC→XBAR with a relationship type that carries no
ratio at all, which is what gates the feature off there; an RTX 2000 Ada and an RTX
3060 Ti both report it that way.

## Testing

RTX PRO 6000 Blackwell Max-Q, driver 610.57.04, under a sustained compute load with
GPC sitting at about 1499 MHz throughout. Clocks are measured with
`CLK_MEASURE_FREQ`, so these are physical rather than requested:

| ratio | GPC | XBAR | measured XBAR/GPC |
|---|---:|---:|---:|
| 0.8999 (factory) | 1498.7 MHz | 1348.4 MHz | 0.900 |
| 1.2 | 1499.1 MHz | 1796.2 MHz | 1.198 |
| 1.0 | 1498.6 MHz | 1492.1 MHz | 0.996 |
| 0.8999 (restored) | 1499.1 MHz | 1348.7 MHz | 0.900 |

The full control block after restore is byte-identical to the preimage.

It is a propagation constraint and not a formula, exactly as #1159 describes. On an
RTX PRO 4000 SFF the same sweep changed nothing, because XBAR there was already
running at about 1990 MHz off its own curve, above GPC times the ratio, so the
relationship was not the binding constraint.

`driver.rs` carries an `#[ignore]`d round-trip test alongside the existing one; run
it with `LACT_TEST_GPU_MINOR=<n> cargo test -- --ignored`. It skips itself when the
variable is unset, and asserts the absence of the control on pre-Blackwell cards.

## What the ratio actually reaches

The name of the relationship is GPC to XBAR, but SYS, Power and Legacy hang off
XBAR through further relationships and follow it. Measuring every domain with
`CLK_MEASURE_FREQ` across the same A-B-A sweep, on the same card and load:

| domain | at 0.9 | at 1.2 | change |
|---|---:|---:|---:|
| XBAR | 1348.3 MHz | 1796.6 MHz | +33.2% |
| SYS | 1413.1 MHz | 1828.2 MHz | +29.4% |
| Power | 1336.8 MHz | 1720.4 MHz | +28.7% |
| Legacy | 1374.5 MHz | 1733.8 MHz | +26.1% |
| GPC | 1498.3 MHz | 1498.7 MHz | 0.0% |
| HUB | 539.9 MHz | 539.8 MHz | 0.0% |
| memory | 13351.8 MHz | 13347.2 MHz | 0.0% |
| Display | 674.9 MHz | 675.2 MHz | 0.0% |

HUB and Display also hang off XBAR but stay pinned, so they do not follow.

Core voltage held at 0.740 V across the whole sweep and board power moved by under
3 W, so the ratio is a clock propagation constraint rather than a voltage control.
At the much higher operating point in #1159 the card did request more MSVDD to
sustain the resulting clocks, which is worth knowing but is a downstream effect.

Because it reaches past XBAR, the control is labelled "GPC Clock Propagation
Ratio" in the UI rather than naming a single destination domain, with an info
popover listing what moves and what does not.

## Caveats

- Same branch-specific caveat as #1158: the layout is private, derived on R610, and
  should not be assumed to hold elsewhere. It is discovered from the info response
  and the sizes are asserted at compile time, so a mismatch should surface as a
  failed control rather than a bad write.
- The driver reports no range for the ratio. The UI is limited to 50–200% as a
  guard rail, chosen because the factory value is 90% and the alternate
  relationships the hardware itself lists span 80% to 200%. Happy to change it.
- Raising the ratio raises the voltage the controller requests for the shared rail,
  so it interacts with the power budget and with the MSVDD ceiling. #1159 measured
  the request going from about 0.923 V to 0.996 V at ratio 1.2.
- The ratio is exposed as an integer percentage, so the factory 0.899902 shows as
  90%. Reset restores the exact factory value rather than 0.90.

Thanks to @Loong0x00 for finding and characterising this in #1159.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Panchovix:
Since noticing that GPC ratio changes more than XBAR, I have renamed the draft/PR, and also added this hint into the UI:

imagen

Panchovix:
Since noticing that GPC ratio changes more than XBAR, I have renamed the draft/PR, and also added this hint into the UI:

imagen

I have the issue about the text alinement.

Panchovix:
Since noticing that GPC ratio changes more than XBAR, I have renamed the draft/PR, and also added this hint into the UI:

imagen

Issue #1151, opened 2026-08-12
Bug Report: LACT does not apply power management settings on AMD Radeon 660M


FauxJD:
### Checklist

- [x] The problem is not listed in the [hardware support matrix](https://github.com/ilya-zlobintsev/LACT/wiki/Hardware-Support) as a known limitation. Examples of known driver-side issues are fans not turning on with a custom curve on RDNA3/4 and missing fan control on laptops.
- [x] I've included a [debug snapshot](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#reporting-issues) if the issue is related to hardware configuration
- [x] I've checked that the issue still happens in the latest [development build](https://github.com/ilya-zlobintsev/lact?tab=readme-ov-file#development-builds).

### Bug description

LACT cannot switch the power management mode (power_dpm_force_performance_level) from auto to manual through the graphical interface. When attempting to apply settings, the following error occurs:

```
Could not apply settings: io error: Invalid argument (os error 22)
Could not commit values to "pp_od_clk_voltage"
```

At the same time:
- Manual switching via terminal works
- After manual switching, frequency control works

System Information

- GPU: AMD Radeon 660M (Rembrandt / YELLOW_CARP)
- Driver: amdgpu
- Kernel parameters: amdgpu.ppfeaturemask=0xffffffff
- LACT version: (please specify)

Steps to Reproduce

1. Open LACT
2. Change any frequency settings (minimum/maximum)
3. Click "Apply"
4. Error appears

Expected Behavior

- LACT should switch the mode from auto to manual
- Frequency settings should be applied
- There should be no errors

Actual Behavior

- LACT cannot switch the mode to manual
- An error occurs when applying settings
- Settings are not applied

What Works (Manual Control)

```bash
# Switching to manual mode (works)
echo "manual" | sudo tee /sys/class/drm/card0/device/power_dpm_force_performance_level

# After this, frequency control works
echo "s 1 400" | sudo tee /sys/class/drm/card0/device/pp_od_clk_voltage
echo "s 2 1899" | sudo tee -a /sys/class/drm/card0/device/pp_od_clk_voltage
echo "c" | sudo tee -a /sys/class/drm/card0/device/pp_od_clk_voltage
```

Available States

```
$ sudo cat /sys/class/drm/card0/device/pp_od_clk_voltage
OD_SCLK:
0:        200Mhz
1:        400Mhz
OD_RANGE:
[...]

Issue #915, opened 2026-02-18
fix: service restart fails to apply clock settings when performance_level is set to manual


hamkido:
### Checklist

- [x] The problem is not listed in the [hardware support matrix](https://github.com/ilya-zlobintsev/LACT/wiki/Hardware-Support) as a known limitation. Examples of known driver-side issues are fans not turning on with a custom curve on RDNA3/4 and missing fan control on laptops.
- [x] I've included a [debug snapshot](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#reporting-issues) if the issue is related to hardware configuration
- [x] I've checked that the issue still happens in the latest [development build](https://github.com/ilya-zlobintsev/lact?tab=readme-ov-file#development-builds).

## Bug Description

When setting the performance level to `manual` and configuring a specific max core/memory clock via the GUI, the settings work correctly at the time of configuration. However, after restarting the `lactd` service (e.g. `systemctl restart lactd`), the saved config fails to apply with an error like "Could not write clocks table".

## Root Cause

In `AmdGpuController::apply_config()` (`lact-daemon/src/server/gpu_controller/amd.rs`), the execution order is incorrect:

1. `reset_clocks_table()` (line 1065)
2. Steam Deck only: set `Manual` performance level (line 1080-1085)
3. **Apply clocks table via `set_clocks_table()`** (line 1087-1118) — requires `Manual` mode
4. **Set performance level to `Manual`** (line 1121-1141) — too late!

On a fresh service start, the GPU defaults to `Auto` mode. Step 3 tries to write to `pp_od_clk_voltage` before `Manual` mode is set in step 4, causing the write to fail.

When the user configures settings through the GUI interactively, the performance level is already set to `Manual` before clock values are written, so it works fine in that case.

## Expected Behavior

The daemon should successfully apply saved clock configuration on service restart when `performance_level: manual` is set in the config.

### System info

```markdown
- LACT version: 0.8.3
- GPU model: amd 680M
- Kernel version: 6.18.2
- Distribution: arch
```

Issue #419, opened 2024-11-21
io error: Invalid argument (os error 22)


numerfolt:
### Checklist

- [X] The problem is not listed in the [hardware support matrix](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#hardware-support) as a known limitation. Examples of known driver-side issues are fans not turning on with a custom curve or inaccurate power limits on RDNA3.
- [X] I've included a [debug snapshot](https://github.com/ilya-zlobintsev/LACT?tab=readme-ov-file#reporting-issues) if the issue is related to hardware configuration

### Bug description

Thank you very much for this tool!

Setting my iGPU clock speed to "automatic", "lowest" or "highest" works, but when I try to set a specific maximum clock speed on my Radeon 780M, I get the below error:

Error Message:
```
2024-11-21T15:24:25.847027Z  WARN update_with_view{input=ApplyChanges component="lact_gui::app::AppModel" id="0x7ffcc2343570"}: lact_gui::app: Could not set power states

Caused by:
0: Got error from daemon, end of client boundary
1: Failed to edit GPU config and set enabled power states
2: Could not apply settings
3: Could not set CoreClock power states
4: io error: Invalid argument (os error 22)
```

Journalctl Output
```
Nov 21 16:19:46 hostname lact[1574]: 2024-11-21T15:19:46.990912Z ERROR lact_daemon::server::handler: could not apply settings: Could not set CoreClock power states
Nov 21 16:19:46 hostname lact[1574]: Caused by:
Nov 21 16:19:46 hostname lact[1574]:     io error: Invalid argument (os error 22)
```

I also cannot set anything except the power limit for my dGPU but probably should open a seperate issue for this?

### System info

```markdown
- LACT version: 0.6.0-release
- GPU model: AMD Ryzen 7 7840HS w/ Radeon 780M Graphics
- Kernel version: 6.8.0-49-generic
- Distribution: KDE Neon
- Framework 16 with dGPU (RX 7700S)
```

Link to the debug snapshot:
https://wolke.numerfolt.de/s/pirjADa2NoAQQKA

Jamgiter:
Same with me. Invalid argument. OS Error 22 when I want to change GPU clock.

Kenzyss:
Same problem. Latest version. Arch, amdgpu 680m

numerfolt:
When I try it with LACT 0.6.0-rlease (commit 3fd7e0e), I get this error message:

```
[...]

PR #1174, opened 2026-08-24
fix: update list of apu devices which need p-levels workaround

makarov-roman:
I decided not to go with libdrm for now. Although it seems better long-term, there might be many nuances  (handle lifecycle, what do should happen when it failed to load,  e.t.c). later we can use existing REQUIRE_MANUAL_DEVICE_IDS as fallback (e.g. for tests). 

refs:
#1151, #915, #419


ilya-zlobintsev (approved):
I do not like needing to have this list hardcoded, but oh well

PR #1175, opened 2026-08-25
fix: simplify amd apply config

makarov-roman:
what I learned:
1. r command requires manual p-lvl on APUs
2. some devices need you to commit after r
3. switching to Auto will restore soft-limits on dGPU and hard limits on APUs.

It wouldn't be amd driver without nuances, right? switching to Auto on APUs will reset min/max core/vram clocks, so unless users will select manual, these values will be overridden. 

knowing that I think the apply can be simplified to:
1. Auto limits reset
2. Manual for OD changes
3. r
5. commit if no clock changes/otherwise commit will be there
...
6. finish with setting p-level to configured value

this flow should work on both APU and dGPUs. But we'd have to re-tests everything.


makarov-roman:
so far tested on steamdeck and rdna3/rdna4

bisspector:
tested on rdna2, lgtm

ilya-zlobintsev (approved):

PR #1178, opened 2026-08-27
feat: oc cards layout

makarov-roman:
image
Goals:

- use screen real estate more efficiently, this is mostly a preparation for contextual stats/charts, but Imo already has better UX/UI
- Unitfy controls/looks. Here a further refactoring needed. I want to make adjustment group/rows more portable. migrate Power block to it, maybe thermals, don't know yet, will experiment later.
- we can add new secondary clocks like sys/crossbar e.t.c as a new block

I'll start from regressions:
1. ~~no OC warning~~ fixed
2. Nvidia OC tooltip. it feels also misleading and a bit outdated. I'm planning to add some description to VRAM offset regarding multiplier.

I also replaced label "you need to set manual p-lvl to use p-state config" with info-dialog to fix the layout, and it's better UX. 

power-state view is temporary, I'll merge them into clocks/ram, but as a separate change, since amd view might be not very simple

there might be such blank space due to how flowbox work. 
image

Before the release I want to make simple masonry layout, like on pic, but up 2 columns
image


lact-gui/src/app/utils/styles.rs
```diff
@@ -64,7 +64,11 @@ pub fn apply_theme(theme: AppTheme) -> anyhow::Result<()> {
         #[allow(deprecated)]
         provider.load_from_data(css);
 
-        style_context_add_provider_for_display(&display, &provider, 900);
+        style_context_add_provider_for_display(
+            &display,
+            &provider,
+            gtk::STYLE_PROVIDER_PRIORITY_SETTINGS,
```
makarov-roman, line 70: adwaita has 200, breeze theme will be 400, and application css 600

ilya-zlobintsev: The current value is set like this because it was broken when there is a custom adwaita override (used in cosmic desktop for example):
https://github.com/ilya-zlobintsev/LACT/pull/931/changes/a3a90fe2b2cc43e5527128ab3f18b90635854c93
https://github.com/ilya-zlobintsev/LACT/pull/931#issuecomment-4016091735

ilya-zlobintsev (commented):
Overall I get the idea and I think it makes sense. The masonry layout should make it more reasonable to have unevenly sized parts of the UI, especially across vendors etc.
A few questions:
- Could we have a proper masonry layout, instead of a flexbox, using [LayoutManager](https://docs.gtk.org/gtk4/class.LayoutManager.html)? It seems hopefully not too complicated, it doesn't need a whole new custom widget, just custom layout manager + Box.
- The double reset button is misleading, because actually everything is reset. Could it be in some global place?

lact-gui/i18n/en/lact_gui.ftl
```diff
@@ -169,21 +172,11 @@ power-mizer-mode-prefer-maximum-performance-description = Favor maximum performa
...
-nvidia-oc-description =
-    Overclocking functionality on Nvidia includes setting offsets for GPU/VRAM clockspeeds and limiting the potential range of clockspeeds using the "locked clocks" feature.
-
-    On many cards, the VRAM clockpeed offset will only affect the actual memory clockspeed by half of the offset value.
-    For example, a +1000MHz VRAM offset may only increase the measured VRAM speed by 500MHz.
-    This is normal, and is how Nvidia handles GDDR data rates. Adjust your overclock accordingly.
-
-    It is possible to achieve a pseudo-undervolt by combining the locked clocks option with a positive clockspeed offset.
-    This will force the GPU to run at a voltage that's constrained by the locked clocks, while achieving a higher clockspeed due to the offset.
-    This can cause system instability if pushed too high.
```
ilya-zlobintsev, line 182: This description is partially outdated, but there is some useful info here (such as VRAM offset not being true to clocks) which is not very clear. We should have this somewhere, or clearly link to the wiki and explain it well there

lact-gui/src/app/pages/oc_page/clocks_frame/mod.rs
```diff
@@ -79,36 +111,27 @@ impl relm4::Component for ClocksFrame {
     type CommandOutput = ();
 
     view! {
-        PageSection::new(&fl!(I18N, "overclock-section")) {
-            set_hide_visible_container: true,
-            add_css_class: "clocks-frame",
+        PageSection::new("") {
```
ilya-zlobintsev, line 114: Very small nitpick: it's been possible to do `PageSection {}` for a while, the `new` is redundant

lact-gui/src/app/pages/oc_page/clocks_frame/mod.rs
```diff
@@ -57,7 +58,38 @@ impl ClocksData {
...
+            | ClockspeedType::VoltageBoost
+            | ClockspeedType::GpuVfCurveClock(_)
+            | ClockspeedType::GpuVfCurveVoltage(_) => Self::Gpu,
+            ClockspeedType::MaxMemoryClock
+            | ClockspeedType::MinMemoryClock
+            | ClockspeedType::MemClockOffset(_)
+            | ClockspeedType::MemVfCurveClock(_)
+            | ClockspeedType::MemVfCurveVoltage(_) => Self::Vram,
+            ClockspeedType::Reset => unreachable!(),
+        };
```
ilya-zlobintsev, line 90: Would be nice to have a `from_clock_type` (or `From` impl) function that exposes this

ilya-zlobintsev (approved):

PR #1179, opened 2026-08-27
fix: set correct p-state configurable value after performance level change

makarov-roman:
I found it when tested #1178

str:
1. amd device/amd snapshot
2. open p-states
3. switch p-level to Manul
4. you'll see that enable p-state configuration checkbox is off, but p-states toggle checkboxes appeared


ilya-zlobintsev (approved):
Will merge after conflicts fix

PR #1193, opened 2026-09-05
feat: oc page QoL

makarov-roman:
some minor changes cherry-picked from #1178 
1. AMD: when you try to enable p-states configuration with non-manual p-level, info dialog will be used to ask to enable it instead of warning label.
2. p-levels/powermizer uses more compact layout with description dimmed text, this is temporary until we "teach" adjustmentRow to handle non-numeric values. 
3. OC Warning was moved to sidebar next to apply button.

### DIFF
--- a/lact-gui/i18n/en/lact_gui.ftl
+++ b/lact-gui/i18n/en/lact_gui.ftl
@@ -151,9 +151,9 @@ performance-level-high-description = Always use the highest clockspeeds for GPU 
 performance-level-low-description = Always use the lowest clockspeeds for GPU and VRAM
 performance-level-manual-description = Manual performance control
 performance-level-profile-standard-description = Fixed profiling mode
-performance-level-profile-min-sclk-description = forces the GPU clock to lowest level
-performance-level-profile-min-mclk-description = forces the VRAM clock to lowest level
-performance-level-profile-peak-description = forces GPU and VRAM clocks to highest levels
+performance-level-profile-min-sclk-description = Forces the GPU clock to lowest level
+performance-level-profile-min-mclk-description = Forces the VRAM clock to lowest level
+performance-level-profile-peak-description = Forces GPU and VRAM clocks to highest levels
 
 performance-level = Performance Level
 power-profile-mode = Power Profile Mode:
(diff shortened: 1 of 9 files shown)


lact-gui/i18n/en/lact_gui.ftl
```diff
@@ -151,9 +151,9 @@ performance-level-high-description = Always use the highest clockspeeds for GPU
 performance-level-low-description = Always use the lowest clockspeeds for GPU and VRAM
 performance-level-manual-description = Manual performance control
 performance-level-profile-standard-description = Fixed profiling mode
-performance-level-profile-min-sclk-description = Profiling mode that forces the GPU clock to lowest level
-performance-level-profile-min-mclk-description = Profiling mode that forces the VRAM clock to lowest level
-performance-level-profile-peak-description = Profiling mode that forces GPU and VRAM clocks to highest levels
+performance-level-profile-min-sclk-description = forces the GPU clock to lowest level
+performance-level-profile-min-mclk-description = forces the VRAM clock to lowest level
+performance-level-profile-peak-description = forces GPU and VRAM clocks to highest levels
```
ilya-zlobintsev: Inconsistent lower/uppercase between the profiles

lact-gui/src/app.rs
```diff
@@ -218,6 +218,26 @@ impl AsyncComponent for AppModel {
...
+
+                                            gtk::Label {
+                                                set_margin_all: 12,
+                                                set_markup: &fl!(I18N, "oc-warning-description"),
+                                                set_xalign: 0.0,
+                                                set_wrap: true,
+                                                set_max_width_chars: 45,
+                                            },
+                                        },
+                                    },
```
ilya-zlobintsev: Do we need to show this all the time? It only applies to the contents of the OC page (and even not all of them). This message might be potentially too scary for some safe operation e.g. reducing power limit

ilya-zlobintsev (commented):
The p-states confirmation dialog is nice

lact-gui/src/app.rs
makarov-roman: well, funnily, my intention was to express the opposite. 
And I believe this is relevant for thermals page as well, in case the cooling will be insufficient.
How about this?

image

ilya-zlobintsev: I don't really agree with this. It might scare new users from using any features, even if they are safe (like the mentioned power limit, or performance level). And I think "What are the risks" is a bit of an abstract wording to be always shown next to a universally used button.

I think if we want to have warnings (e.g. overheating with a bad fan curve), it should be targeted (like on fan page), rather than global

makarov-roman: okay, removed. 
most of it is common sense anyway.

ilya-zlobintsev (approved):

PR #1194, opened 2026-09-05
refactor: adjustment-row on oc page

makarov-roman:
I have another branch where I migrated Thermal page to adjustment row, but with the layout changes the diff swiftly grew. And Fan controls has complex state, so let's check them in isolation.

the goal of this refactoring - I want to have universal component which can control Adjustment value on any page, so we have unified behaviour (like remember we missed the fix with scrollbar on thermal page because it had different input?).

later I want  Adjustment row to be able to handle bool/select


lact-gui/src/app/components/adjustment_row.rs
```diff
@@ -0,0 +1,215 @@
...
+use crate::app::utils::ext::make_event_controller_no_scroll;
+use adw::prelude::*;
+use relm4::{FactorySender, RelmWidgetExt, css, factory::FactoryComponent};
+use std::marker::PhantomData;
+
+pub struct AdjustmentRow<Key> {
+    title: String,
+    title_tooltip: String,
+    info_text: String,
+    _key: PhantomData<Key>,
```
ilya-zlobintsev, line 11: Should this actually store the key, rather than PhantomData? We have it available as `index` in `init_model`

lact-gui/src/app/pages/oc_page/clocks_frame/mod.rs
```diff
@@ -38,6 +36,39 @@ pub struct ClocksFrame {
...
+
+impl Default for ClocksData {
+    fn default() -> Self {
+        Self {
+            current: 0,
+            min: 0,
+            max: 0,
+            custom_title: None,
+            is_secondary: false,
+            step: 10,
```
ilya-zlobintsev: It's not introduced by this PR, but I've just noticed that on AMD, the voltage offset step is 10 mV. I think it would be better if it was 5 or 1 (Nvidia voltage boost % step is already 1)

lact-gui/src/app/components/adjustment_row.rs
makarov-roman: We don't reference the key anywhere, we can easily add it later if needed

lact-gui/src/app/pages/oc_page/clocks_frame/mod.rs
makarov-roman: oh yeah. 
let's use 1 mV for voltages, 1% for percentages, 5Mhz for clocks
When I did OC last time - the stepping was very annoying due to how snapping works.

ilya-zlobintsev (approved):

PR #1201, opened 2026-09-11
fix: restore power state configuration on GUI startup

Xarth-Mai:
Reopening the GUI leaves **P-State Configuration** unchecked even when the daemon still reports a saved selection such as `memory_clock: [3]`. Reproduced with matching development GUI and daemon builds, including #1179, on an RX 6650 XT

The parent queues updates to the child lists, then checks their contents before they have populated. This clears the restored toggle. Use availability from the incoming power states for configurability and list visibility instead

Validation:
- The regression test fails on the original code and passes with the fix, including switching to a device without power states
- Reopening the patched GUI restores the toggle; the saved configuration is unchanged. Hardware verification did not click Apply
- Build, formatting, GUI tests (all features), and GUI Clippy passed

The GTK test requires a display and is ignored by default. Run it with:
```sh
cargo test -p lact-gui --features gtk-tests restores_configured_states_before_child_lists_update -- --ignored --test-threads=1
```

### DIFF
--- a/lact-gui/src/app/pages/oc_page/power_states/power_states_frame.rs
+++ b/lact-gui/src/app/pages/oc_page/power_states/power_states_frame.rs
@@ -29,6 +29,8 @@ pub struct PowerStatesFrame {
     performance_level: Option<PerformanceLevel>,
     configured_signal: SignalHandlerId,
     vram_clock_ratio: f64,
+    has_core_states: bool,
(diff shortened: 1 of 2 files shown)


Xarth-Mai (commented):
Passed the test.

ilya-zlobintsev (approved):

PR #1203, opened 2026-09-13
feat: detachable pages

makarov-roman:
Detachable windows serve the case when you want to see multiple pages. It's already addressed with graphs and processes being a dialogs rather than page, but makes the structure overall more complex.

image

implementation wise it's rather simple, detachable_page creates a hidden empty window for each page, and when we toggle detachment, it moves the widget subtree between main window and detached.

it's draft because I want you to review the core part. After that I want to make 2 changes:
1. save per-page detached state into UI config, similar to what we do with selected page
2. move show graphs and processes into sidebar as pages

### DIFF
--- a/lact-gui/i18n/en/lact_gui.ftl
+++ b/lact-gui/i18n/en/lact_gui.ftl
@@ -3,6 +3,10 @@ oc-page = Overclocking
 thermals-page = Thermals
 software-page = Software Info
 displays-page = Display Info
+detach-page = Open {$page} in a Window
+page-detached = This page is open in another window.
+show-page-window = Show Window
+reattach-page = Return to Main Window
 
 hardware-info = Hardware Information
 
--- a/lact-gui/i18n/en/lact_gui.ftl
+++ b/lact-gui/i18n/en/lact_gui.ftl
@@ -3,6 +3,7 @@ oc-page = Overclocking
 thermals-page = Thermals
 software-page = Software Info
 displays-page = Display Info
+sidebar = Sidebar
 detach-page = Open {$page} in a Window
 page-detached = This page is open in another window.
 show-page-window = Show Window
--- a/lact-gui/src/app.css
+++ b/lact-gui/src/app.css
@@ -14,6 +14,15 @@
     margin: 4px 8px 8px 8px;
 }
 
+.page-detach-button {
+    opacity: 0;
+}
+
+.navigation-sidebar row:hover .page-detach-button,
+.page-detach-button:focus-visible {
+    opacity: 1;
+}
+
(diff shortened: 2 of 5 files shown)


ilya-zlobintsev:
I definitely like the idea, and I think it makes sense to move the graphs and process monitor in as pages for discoverability (though let's keep that for a separate PR).

Some ideas:
- Could the windows be created dynamically on demand, rather than always there, just hidden? It's not worth it if this ends up being more complex code wise, but would feel "cleaner" in terms of widget lifetime
- The detach button is very easy to miss right now, maybe it makes sense to have it always visible for the currently selected tab, or alternatively have it in the header
- I'm not sure if it's intended for the detached windows to always be above the main one

makarov-roman:
>  Could the windows be created dynamically on demand, rather than always there, just hidden? It's not worth it if this ends up being more complex code wise, but would feel "cleaner" in terms of widget lifetime

I just want to avoid complexity of handling it's lifecycle, with just few pages it's not noticeable. 

>  The detach button is very easy to miss right now, maybe it makes sense to have it always visible for the currently selected tab, or alternatively have it in the header

I don't think it's easy to miss it. When you switching the page, you're looking at the sidebar, at the button you're clicking. Which won't be the case with the header.  

>  I'm not sure if it's intended for the detached windows to always be above the main one

it's not, It was a side effect of set_transient_for. fixed

makarov-roman:
then we can do config part also via separate PR

ilya-zlobintsev:
> I don't think it's easy to miss it. When you switching the page, you're looking at the sidebar, at the button you're clicking. Which won't be the case with the header.

If you aren't deeply familiar with the UI, there are likely elements you are not actively looking at. At least my feeling is that it will be very easy to miss something that only briefly shows up when hovering

makarov-roman:
It might be UI philosophy question. Generally our perception is good at seeing the difference. So popping icon on the arguably most used element during hover is noticeable. 

It's different when you know the UI and want to find the detach button (like you're now), but when you need another page, you're switching to it, and you see the icon popping - it probably even more "attractive" to our attention than a static icon.

ilya-zlobintsev (commented):
I'm not sure it's intended that the buttons are taller now:

Image

Image

ilya-zlobintsev:
> It's different when you know the UI and want to find the detach button (like you're now), but when you need another page, you're switching to it, and you see the icon popping - it probably even more "attractive" to our attention than a static icon.

Would always showing the button on current page + when hovering on others cover both cases? It would still be something that "pops" when navigating so you notice it, but still findable in the UI when not interacting

ilya-zlobintsev (approved):

PR #1205, opened 2026-09-16
feat(nvidia): add opt-in experimental power control

Panchovix:
Add a per-GPU opt-in for experimental power caps below the VBIOS minimum on compatible drivers, with a visible warning. NVML remains the default; ioctl mode applies all caps and resets without automatic backend switching, while preserving the native maximum.

Validated with backend/schema tests and a GTK opt-in, profile-loading, reset and opt-out test.

imagen


ilya-zlobintsev:
I don't like this idea. Power cap is a basic feature that is expected to always work (and safely), automatically switching from the official NVML calls to undocumented queries in order to circumvent intended hardware limits is absolutely not something that should happen.

At most, this could be a separate opt-in that's clearly labeled with a warning for changing how the power cap gets applied, where enabling it would switch to the ioctl path. But never something that changes how the setting is applied based on the value provided.

Panchovix:
Perfect, I will look a way to do as an optional opt in.

Panchovix:
Pushed a new commit that makes it opt-in, tested and working without issues, only tested on nvidia 610.57.04, but I'm gonna test with 615.71.09 now.

EDIT: Works fine on 615.71.09 as well.

imagen

imagen

imagen

makarov-roman:
hi.
I assume it might be relevant to local llms inference. 
Do you have any real world example where going below bios min limit makes sense?
code is liability, it should be tested, maintained, so I'm trying to understand what are the practical benefits.

Isn't it better and enough to use undervolt(and maybe locked clocks) to increase efficiency?

Panchovix:
I'm not home right now for a given test, but as you say, it is relevant mostly in 3 cases:

- Local LLM inference: power limit barely affects text generation performance, and having a higher "min" power limit makes it less efficient that stock.
- A very small subset of users do shunt mod their cards (1.5X, 2X, etc). Nvidia on Blackwell did put the min power limits quite higher, so in the example on a RTX 6000 PRO MaxQ, min VBIOS power limit is 250W. At stock that's already too high and with a shunt that can be 400-500W. With this PR you can basically control to I.e. 350W (175W limit) reported, which is higher than the min of the stock VBIOS. Let me know if I explained myself correctly here, English is not my first language.

And finally the 3rd one of something you mentioned. I indeed first went the classic way to undervolt by setting a locked core clock and then an offset. The issue is:
- Depending of the task, it can use a given amount of power in one but more or less in another one.
- If your min power limit is still higher that the power you want with an undervolt, it can still go higher than the amount you want (again, let me know if this is understandable).
- Doing a overclock + power limit let you reach higher clocks on multipart tasks. For example, if you do image diffusion, a part of the generation can use less power than other (upscaling, decoding). Having a max core clock set there you can lose performance vs having a higher clock plus a power limit, since it would use higher clocks in parts where it doesn't need much power, and power limit when it needs.

Basically those are the reasoning for me to do this PR. Now it is understandable that it isn't a must, so if it has to be closed, no issues.

danii:
I have been using this patch for the past few days, it has been working well for me. My understanding is that locking clocks could make the card use more power than necessary, whereas setting a power limit lets the card find the highest clocks it can maintain within that power limit. That's one of my reasons for using this PR, that and it's much more convenient than fiddling with the clocks to get the desired "effective power limit".

phazei:
Amazeballs!  Would love to see if this is at all applicable to windows

PR #1207, opened 2026-09-17
feat: update adjustment row UI

makarov-roman:
1. move units into subtitle
2. display values range in subtitles
3. popover is triggered by text, rather than icon (which looks differently on each DE)
image


ilya-zlobintsev (commented):
I like this idea, but the legibility of the range is not very good, especially for negative values:

Image

You really need to look carefully here to see the -200 apart from the separator dot

makarov-roman:
image

makarov-roman:
I also decided to use boxed-list to make each frame more consistent with adwaita structure wise. Maybe it will look better if somebody decide to theme adwaita. And I don't have to handle separators.

ilya-zlobintsev (approved):

PR #1208, opened 2026-09-18
refactor: migrate thermals-page to adjustmentRow

makarov-roman:
sorry for another refactoring, but It's necessary to make things right with Reset/Default change we discussed. I want to replace adjustmentValue with new generic entity for cleaner is_changed,is_default, get_changed, load and reset apis, also I want it to handle bool/strings so we can use it for all the settings we persist in gpu config.

And AdjustmentRow will be an abstraction for numeric type, which will use new adjustmentValue inside. 

This PR shouldn't have any functional changes. we're getting:
1. same UI same behaviour as for OC page
2. ability to change adjustmentValue implementation without touching thermals

I've checked that the same options are exposed for amd/nvidia as on master


makarov-roman:
The way it looks right now is not perfect.
Ideally I want to move thermalPage to OC page as a thermalFrame. fan mode should become a row. Curve would be a clickable row with external window.

Making this change right now would be hard to review. Switching to adjustment row is a step towards that.

